PR #47, opened 2020-06-22
get station data in interface

jannisvisser:


jannisvisser:
@elwinschmitz I need to manually run 'npm run fix:syntax' within this setup to correct prettier styling (which will otherwise lead to failed tests). Not sure if that's how it's supposed?

interfaces/overview/src/app/components/map/map.component.ts
```diff
@@ -8,6 +9,7 @@ import { Layer, MapService } from 'src/app/services/map.service';
 })
 export class MapComponent implements OnInit {
   public layers: Layer[];
+  public stations: Station[];
```
elwinschmitz, line 12: I wonder if there will be only 1 set of 'stations' on the map.
These stations will probably be in "a point layer of (water-)stations".

So the MapService needs to provide a (new) layer with these stations as its contents..
So something like:

```
// this could happen in some `ngOnInit`-step, or triggered by some user-event:
const stations = this.apiService.getStation();
const stationsLayer = this.mapService.createLayer('point', stations);
this.mapService.addLayer(stationsLayer);
```
Or these last steps could be combined differently as:
```
this.mapService.addPointLayerFromArray(stations);

// and the contents of that 👆 would be:
const layer = this.createLayer('point', input);
this.addLayer(layer);
```

interfaces/overview/src/app/components/matrix/matrix.component.spec.ts
```diff
@@ -9,7 +10,7 @@ describe('MatrixComponent', () => {
   beforeEach(async(() => {
     TestBed.configureTestingModule({
       declarations: [MatrixComponent],
-      imports: [IonicModule.forRoot()],
+      imports: [IonicModule.forRoot(), HttpClientModule],
```
elwinschmitz, line 13: See other comment above.

interfaces/overview/src/app/services/jwt.service.ts
```diff
@@ -0,0 +1,26 @@
+import { Injectable } from '@angular/core';
+import { JwtHelperService } from '@auth0/angular-jwt';
```
elwinschmitz, line 2: I think we could use this package even more, so we don't need to handle the headers in the ApiService and we don't need this whole JWT-Service.
https://github.com/auth0/angular2-jwt#usage-injection

But it might be more work to refactor this then it saves at the moment...

elwinschmitz:
> @elwinschmitz I need to manually run 'npm run fix:syntax' within this setup to correct prettier styling (which will otherwise lead to failed tests). Not sure if that's how it's supposed?

If you've generated files or copied them from elsewhere without editing then it makes sense, but I think it is still related to your Windows line-endings issue.

I think we need to look into setting up this repository in the right way (with a `.gitattributes`-file) and then do a clean checkout on your machine to 'fix' the local index/config or something.
If that doesn't fix it, I don't really see what the differences are here in this repo compared to the 121-repo/set-up...

PR #49, opened 2020-06-23
Docker compose for CI

gulfaraz:
1. use dotenv file for secrets and config
2. add dockerfile for web interface
3. create docker compose for ci
4. use nginx to serve app (only for interface)


jannisvisser:
@gulfaraz is this code also tested to work as Azure web-app already? Or do you first need to merge before you can test that further?

gulfaraz:
@jannisvisser these changes are to allow us to setup the azure web app via docker. I haven't tested these as the web app. I will test on the feature branch before merging into master.

PR #67, opened 2020-07-01
Deployed to Azure

gulfaraz:
https://ibf-system.azurewebsites.net/


docker-compose.azure.yml
```diff
@@ -0,0 +1,27 @@
+version: "3.8"
+
+services:
+    ibf-api-service:
+        image: gulfaraz/ibf-api-service
```
elwinschmitz, line 5: Should we create a "rodekruis"-account on DockerHub for this?
So that we're not depending on a personal account for this.

elwinschmitz:
So for my understanding;

- the `docker-compose.yml`-file describes the "local development environment" only, right?
  - it exposes the API-service from the 'internal' port 3000 to the 'docker/external' locahost:80
  - it 'proxys' the Angular dev-server via port 4200 (running in dev-mode)

- the docker-compose.azure.yml`-file is the "production deployment environment", right?
  - it exposes the API-service at <https://ibf-system.azurewebsites.net/api/> (and /docs)
  - it serves all static files from the `/dist/`-folder on <https://ibf-system.azurewebsites.net>  
    When do these files get generated? By the `/overview/Dockerfile`? Where is it configured how that is run/created/invoked?

Because when I look at the (Angular)interface currently running on <https://ibf-system.azurewebsites.net>, it is not a production-build (it says so in the console.)? Is that because this PR isn't merged yet?

gulfaraz:
> * the `docker-compose.yml`-file describes the "local development environment" only, right?

`docker-compose.yml` describes for manual prod envs
  
>   * it exposes the API-service from the 'internal' port 3000 to the 'docker/external' locahost:80

Yes

>   * it 'proxys' the Angular dev-server via port 4200 (running in dev-mode)

No. The `ibf-web-interface` service is called only to build the app and place the built files in the `api-web-interface-root` volume. This service terminates after the files are built. This volume is used by the `ibf-api-service` to serve the frontend.

> * the `docker-compose.azure.yml`-file is the "production deployment environment", right?

`docker-compose.yml` describes for prod envs which pull from docker hub (ideal for auto deploys like Azure)

>   * it exposes the API-service at https://ibf-system.azurewebsites.net/api/ (and /docs)

Yes

>   * it serves all static files from the `/dist/`-folder on https://ibf-system.azurewebsites.net

Yes

>     When do these files get generated? By the `/overview/Dockerfile`? Where is it configured how that is run/created/invoked?

The `dist` files are generated [here](https://github.com/rodekruis/IBF-system/blob/build.azure_webapp/interfaces/overview/Dockerfile#L19) when building the image

> Because when I look at the (Angular)interface currently running on https://ibf-system.azurewebsites.net, it is not a production-build (it says so in the console.)? Is that because this PR isn't merged yet?

Yes, I see the message. How do I switch to prod build? Is it a env variable? Currently, I'm using `npm run build --prod`

elwinschmitz:
@gulfaraz  Aha, so the `ibf-web-interface` service is just a container to create the build-artifact "`dist`" in. As a build-tool-dependency for the API-service. 👍 

Regarding the "production mode"-build;
Maybe it could make a difference to use:
```
npm run build -- --prod
```

As the npm-script "build" doesn't _really_ accept arguments, but when you "pass them along" with the ` -- `-thing, it works slightly different for `--help`, `--prod`-style flags then 'single word' command line parameters....

elwinschmitz:
@gulfaraz Ah, now that I think of it, maybe we need to set some ENV-variables for the interface... But I think we can do that in a separate next step as well..

docker-compose.azure.yml
gulfaraz: updated to `rodekruis` images.. I've sent team invites on docker hub for access..

gulfaraz:
> ```
> npm run build -- --prod
> ```

This worked

gulfaraz:
I'm unable to setup [automated builds](https://docs.docker.com/docker-hub/builds) for these images. The rodekruis repos are not listed at step 4 of the [configuration steps](https://docs.docker.com/docker-hub/builds/#configure-automated-build-settings).

PR #85, opened 2020-07-07
add context to docker, fix dependencies in SQL, update README

jmargutt:
small bug fixes to run the pipeline @jannisvisser please review

### DIFF
--- a/services/IBF-pipeline/Dockerfile.template
+++ b/services/IBF-pipeline/Dockerfile.template
@@ -28,6 +28,8 @@ WORKDIR /home/ibf/pipeline/
 # install dependencies
 COPY pipeline/requirements.txt /home/ibf/pipeline/
 RUN pip install -r requirements.txt
+ADD pipeline /home/ibf/pipeline
+ADD geoserver /home/ibf/geoserver
 
 # set up cronjob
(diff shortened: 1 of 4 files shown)


jannisvisser:
I get that the ${PWD} may be OS-specific and may therefore not work, but can you simply remove it? In my experience you need to specify what local folder you are connecting to a volume. Which is in this particular case needed/used to get your local code live in the container as well. So if you make changes to your local code, it's immediately changed in the container as well. @jmargutt

jmargutt:
@jannisvisser it makes sense, how would you explain it clearly in the readme?

jannisvisser:
@jmargutt something like: 

1. Build Docker image (from the IBF-pipeline root folder) and run container with volume. ${PWD} takes automatically your Present Working Directory as the local folder to attach the volume though. ${PWD} may be Windows and/or PowerShell-specific. If this fails, you can always replace it by the literal path (e.g. "C:/IBF-system:/home/ibf" instead of "${PWD}:/home/ibf").
```
docker build . -t ibf-pipeline
docker run --net=host --name=ibf-pipeline -v ${PWD}:/home/ibf --restart always -it ibf-pipeline
```

PR #111, opened 2020-07-24
Edits pipeline for EGY

p-phung:


p-phung:
updated requirements
updated dockerfile
built docker image
tested with UGA, ZMB, and EGY

PR #116, opened 2020-07-27
feat: timeline and setup layer subscribers

gulfaraz:
https://dev.azure.com/redcrossnl/IBF%20System/_workitems/edit/2639/

### DIFF
--- a/interfaces/overview/package-lock.json
+++ b/interfaces/overview/package-lock.json
@@ -10057,6 +10057,11 @@
       "integrity": "sha512-vVqVZQyf3WLx2Shd0qJ9xuvqgAyKPLAiqITEtqW0oIUjzo3PePDd6fW9iFz30ef7Ysp/oiWqbhszeGWW2T6Gzw==",
       "dev": true
     },
+    "moment": {
+      "version": "2.27.0",
+      "resolved": "https://registry.npmjs.org/moment/-/moment-2.27.0.tgz",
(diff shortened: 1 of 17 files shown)


jannisvisser:
@gulfaraz looks good! some small points:
1. I guess the current red shades are just dummy and don't mean anything right?
2. the 'today' button is now also clickable. I don't think it should be, as there isn't any meaning to it. It should be styled differently than the rest though.
3. for Uganda at the moment there is only a 7-day forecast, not a 3-day one. So ideal would be to make this dynamic based on country-settings. But for now maybe just turn it off..
4. I don't think it's optimally clear yet, which settings is currently selected. But I get what you're doing, and don't know how to improve at the moment.

jannisvisser:
@gulfaraz and given that 'today' is diasbled, and '3-day' as well, then '7-day' should also by default be color-filled as the currently visualized setting.

gulfaraz:
> 1. I guess the current red shades are just dummy and don't mean anything right?

Currently, the red buttons are only to demonstrate the visual difference. We can configure which buttons are red based on the API response.

> 2. the 'today' button is now also clickable. I don't think it should be, as there isn't any meaning to it. It should be styled differently than the rest though.

Done

> 3. for Uganda at the moment there is only a 7-day forecast, not a 3-day one. So ideal would be to make this dynamic based on country-settings. But for now maybe just turn it off..

I've turned it off for now.

> 4. I don't think it's optimally clear yet, which settings is currently selected. But I get what you're doing, and don't know how to improve at the moment.

The selected button is indicated by a solid fill.

>and given that 'today' is diasbled, and '3-day' as well, then '7-day' should also by default be color-filled as the currently visualized setting.

Done

PR #123, opened 2020-08-03
get alert dynamically by day in timeline

jannisvisser:


jannisvisser:
@gulfaraz this is the PR I just mentioned. I circumvented the development-database in this case by making an additional view (subscripted with _NEW) instead of replacing the old one. This way both local app and azure app will work. But of course I need to remove the old view and the _NEW subscript at some later stage, so this still sucks..

Some final notes I'm noticing just now while looking. Mentioning it here because I wouldn't know where else. Feel free to do something with it if you have spare time ;)
- 3-day is not actually a lead-time for Uganda, it is for Zambia. That should be dynamic.
- For our own sanity, we could make a country-switch (in the header-bar), even if it's not on the backlog directly.
- It seems that the admin-region layer is reloaded when you switch between 3- and 7-day, while it is static data. That should be improved..

PR #155, opened 2020-09-11
AB#3158 email replace text

jannisvisser:


jannisvisser:
@gulfaraz I continued work with this branch. Quite far, but no completely there yet:
- different logo based on country
- the table is looking quite ugly like this (rows need to be styled)
- what should (ALERT-LEVEL) on bottom of email be, if we already have that in table as well.
- Does the unsubscribe link actually work like this? 
- ...?

PR #212, opened 2020-11-09
AB#4323 update webhook script

jannisvisser:
@gulfaraz 
- this updates the webhook-script to allow for the difference between when they are deployed
- in line with the script I have already changed the webhook-service on "IBF-system" to NODE_ENV=staging ("IBF-system-test" has NODE_ENV=test)
- I have already switched around the github-webhooks as well. "IBF-system" is triggered on release, "IBF-system-test" on PR.
- the readme+changelog is copied/adapted from 121


jannisvisser:
@gulfaraz the above message is now outdated, as we just "turned back" bullet 2 and 3 during our conversation.

gulfaraz:
I see some overlap with #216

PR #235, opened 2020-11-13
AB#4236 easy fixes 13-inch screen user-test

jannisvisser:


jannisvisser:
@gulfaraz please have a look if in your professional opinion I didn't go overboard with decreasing font-sizes/margin/padding..

gulfaraz:
@jannisvisser The changes look reasonable. I've listed my opinion to be aware of,

1. The font size is readable to me but I wonder if it's too small for the general audience. The smaller font size gives the impression that we're trying to fit as much information as possible on one screen. This indicates poor UX and no structure in the presented information. I recommend spending time with design to prioritize information to be shown and expand further based on the user interest/interaction.
2. In the aggregates component, is it important to make a distinction between the vulnerability and exposure layers? If not, we can remove the group heading bar.

jannisvisser:
@gulfaraz 
1. I agree. But for the short-term should we increase the font-size a bit (to 13px maybe instead of 12px. it was 14px before.) 
2. I've already removed the group heading bar right? Or do you mean also removing the spacing in between? Because that I think is good to keep for now.

gulfaraz:
1. Let's leave the font size and see if someone requests the increase. 
2. Okay, let's leave the gap.

PR #257, opened 2020-11-16
AB#4475 support microsoft edge

jannisvisser:


jannisvisser:
@gulfaraz see some changes I made for Microsoft Edge support
- only the timeline-height is not working properly yet. The 100% !important is needed, but the height is now too big in Edge. When I unclick+reclick the height checkbox in Developer Tools, it does suddenly adjust OK..
- I had some conflict errors to solve, which indicated you recently worked on aggregates/area-of-focus-summary as well. That was not related to this same point or something right?
- It sucks that in general we have to keep track of features that are not supported in Edge (or in general in all browsers). What is your meta-view on how we should handle that?

jannisvisser:
@gulfaraz also the DOMrect stuff in the layer-info-popup was not supported in Edge, so I left that out, which means the popup shows centrally in the screen, which is fine I think.

PR #316, opened 2021-01-04
AB#5405 text black if trigger

jannisvisser:


jannisvisser:
@gulfaraz see this change

interfaces/IBF-dashboard/src/app/pages/dashboard/dashboard.page.scss
```diff
@@ -66,7 +66,8 @@ app-export-view {
           .component-title-thumbnail {
             height: 24px;
             width: 24px;
-            display: none;
+            display: block;
+            background-image: url('../../../assets/icons/alert-white.svg');
```
gulfaraz, line 70: @jannisvisser do we need to show a warning icon if there is no trigger?

interfaces/IBF-dashboard/src/global.scss
```diff
@@ -57,15 +57,18 @@
...
       background-color: var(--ion-color-ibf-salmon) !important;
 
-      .component-title-item {
+      .component-title-item,
+      .today-button {
+        color: var(--ion-color-ibf-black) !important;
+
         .component-title-thumbnail {
           display: block !important;
+          background-image: url('assets/icons/alert-black.svg') !important;
```
gulfaraz, line 71: @jannisvisser this url is different in dashboard.page.scss, i.e.) '../../../' is missing here. Is this difference intended as a test?

jannisvisser: no, because the respective css files are in different places, this was actually how it was needed to work

interfaces/IBF-dashboard/src/app/pages/dashboard/dashboard.page.scss
jannisvisser: @gulfaraz good point.. I didn't think about it, just turned it from white to black.. I'll remove it then.

PR #330, opened 2021-01-08
AB#5045 Improve readability layer names

jannisvisser:


jannisvisser:
@gulfaraz see the minor (unrefined) change I made to separate the layer-overview from the timeline-component visually. I thought this would maybe solve Orla's issues with that, and I think it looks better personally as well. But it's unplanned, so feel free to turn back if you disagree.

PR #344, opened 2021-01-11
AB#5527 feat: show exposed pop. by default if trigger

jannisvisser:
@gulfaraz Is this commit message correct for Conventional Commit?

### DIFF
--- a/interfaces/IBF-dashboard/src/app/services/map.service.ts
+++ b/interfaces/IBF-dashboard/src/app/services/map.service.ts
@@ -97,7 +97,11 @@ export class MapService {
       label: IbfLayerLabel[indicator.name],
       type: IbfLayerType.shape,
       description: 'loadAggregateLayer',
-      active: indicator.active,
+      active:
+        indicator.name === IbfLayerName.population_affected &&
+        this.eventService.state.activeTrigger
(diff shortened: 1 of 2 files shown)


gulfaraz:
I'm not sure if the tools pick up the type of commit if it's in between. We can test it but I suggest using one of the following to be safe,

`feat: AB#5527 show exposed pop. by default if trigger`

`feat: show exposed pop. by default if trigger AB#5527`

interfaces/IBF-dashboard/src/app/services/map.service.ts
```diff
@@ -97,7 +97,11 @@ export class MapService {
       label: IbfLayerLabel[indicator.name],
       type: IbfLayerType.shape,
       description: 'loadAggregateLayer',
-      active: indicator.active,
+      active:
```
gulfaraz, line 100: @jannisvisser shouldn't the API send the correct value for active instead of changing this in the dashboard?

jannisvisser: @gulfaraz sure. you mean that also the trigger-boolean is checked in API-service? Feel free to adapt.

PR #411, opened 2021-01-29
fix: remove [0,0] waterpoint locations AB#5846

jannisvisser:


services/API-service/src/api/waterpoints/waterpoints.service.ts
```diff
@@ -40,6 +40,11 @@ export class WaterpointsService {
         .get(path, { headers: this.headers })
         .subscribe((response): void => {
           const result = response.data;
+          result.features = result.features.filter(
```
gulfaraz, line 43: For Uganda, `https://data.waterpointdata.org/resource/amwk-dedf.geojson?$where=water_source is not null AND water_source !='Lake' AND within_polygon(location, 'POLYGON ((35.00105362800008 -1.48147434699996, 35.00105362800008 4.231367005000038, 29.57216169200007 4.231367005000038, 29.57216169200007 -1.48147434699996, 35.00105362800008 -1.48147434699996))')&$limit=100&status_id=yes&country_id=UG`

@jannisvisser We should avoid performing the filter on our server when there is an option to offload that. It's easier to send a country bounding box as a filter parameter.

jannisvisser: @gulfaraz I tried this, but it kept failing without clear error message. But go for it then. (But avoid spending too much time on it, it's not worth it..)

gulfaraz: @jannisvisser We should spend some time on this now so we don't have to come back to this in the future. These sort of sub-optimal implementations add up and lead to a load time of 30 seconds, especially so when on the backend where everyone's requests are served by the same CPU.

PR #421, opened 2021-02-01
fix: waterpoints api should use polygon filter

gulfaraz:


gulfaraz:
AB#5846

jannisvisser (approved):
@gulfaraz checked this locally and works.

PR #424, opened 2021-02-01
fix in loading aggregates + fix in deploy-script

jannisvisser:


tools/deploy.sh
```diff
@@ -67,7 +67,7 @@ function deploy() {
         do
             echo "$SCHEMA"
             rm -f tools/db-dumps/ibf_$SCHEMA.dump
-            PGPASSWORD=$DB_PASSWORD pg_dump -U $DB_USERNAME -Fc -f tools/db-dumps/ibf_$SCHEMA.dump -h $DB_HOST -n \"$SCHEMA\" $DB_DATABASE
```
jannisvisser, line 70: @gulfaraz I changed this back. The point here is that it is dumping 
- from **geonode_datav3**
- to **$DB_DATABASE** (that being 'test' or 'staging' depending on which server the script is run on)

gulfaraz: so **geonode_datav3** is like a temp database?

jannisvisser: @gulfaraz no, it's the development database

PR #450, opened 2021-02-09
Ruben fix pipeline pip AB#5968

RubenGeo:


services/IBF-pipeline/pipeline/runSetup.py
```diff
@@ -2,6 +2,7 @@
 from lib.setup.createSubFolders import createSubFolders
 from lib.setup.staticDataDb import uploadStaticToDb, processStaticDataDb
 from settings import *
+from secrets import *
```
gulfaraz, line 5: @jannisvisser is the `runSetup.py` file still needed? If not, can we delete the file?

services/IBF-pipeline/pipeline/runCron.py
```diff
@@ -6,6 +6,7 @@
 import time
 import datetime
 from settings import *
+from secrets import *
```
gulfaraz, line 9: @RubenGeo I get this error when running `runCron.py`

Screenshot 2021-02-09 at 13 38 20

services/IBF-pipeline/Dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
```
gulfaraz, line 1: Works for now but would be nice to upgrade to 20.04 (if possible).

services/IBF-pipeline/pipeline/runSetup.py
gulfaraz, line 5: @jannisvisser is the runSetup.py file still needed? If not, can we delete the file?

gulfaraz (changes requested):

PR #500, opened 2021-02-22
fix: trigger country-subscription at refresh AB#6272

jannisvisser:


jannisvisser:
@RubenGeo FYI: this was the solution (to trigger country-subscription also at general 'initialization' instead of only at login)

PR #588, opened 2021-03-18
fix: populationAffected column should not get dropped on synchronize …

gulfaraz:
…[AB#6690](https://dev.azure.com/redcrossnl/056030a3-96d2-4070-8718-48f602a2af75/_workitems/edit/6690)


gulfaraz:
@jannisvisser @RubenGeo please test on your local before merging. If you get the column contains null values error, please don't set `nullable: true` instead run the pipeline again which will populate the right value into the column.

jannisvisser:
@gulfaraz thanks for noticing this. I wasn't aware that this table was created through an entity (while being in the IBF-pipeline-output schema). As such I removed the DDL create table statement completely again, as it's not needed and will lead to bugs again with more changes in the future.

PR #625, opened 2021-03-23
fix: cypress file location2

RubenGeo:


gulfaraz:
@RubenGeo it appears this PR broke the lighthouse setup

`/var/tmp/ibf-2021-03-25.stderr.log` on the test VM has
```
ENOENT: no such file or directory, open '/home/ibf-user/tests/lighthouse/lighthouserc.js'
```

RubenGeo:
I will what happen if i remove the cd ..

PR #626, opened 2021-03-23
fix: use observables instead of promises AB#6739

gulfaraz:


interfaces/IBF-dashboard/src/app/services/map.service.ts
```diff
@@ -437,189 +438,104 @@ export class MapService {
...
+        .getWaterPoints(this.country.countryCodeISO3)
+        .pipe(shareReplay(1));
     } else if (
       layer.name === IbfLayerName.redCrossBranches ||
       layer.name === IbfLayerName.redCrescentBranches
     ) {
-      data = await this.getRedCrossBranches();
+      layerData = this.apiService
+        .getRedCrossBranches(this.country.countryCodeISO3)
+        .pipe(shareReplay(1));
```
jannisvisser, line 509: @gulfaraz what does .pipe(shareReplay(1)) exactly do?

jannisvisser (approved):
@gulfaraz I have checked this out locally, and can confirm that everything still works.
- all Promises left are all Promise<void> type and are not necessary to remove as well?
- see one specific question here about .pipe(shareReplay(1))

But the answers are no reason to not already merge, so doing that.

PR #632, opened 2021-03-26
fix: move exposed pop. layer active if trigger to front-end AB#6728

jannisvisser:


jannisvisser:
@gulfaraz suggesting this as solution. do you approve?

gulfaraz:
@jannisvisser what's the reasoning behind this? It appears to treat a symptom of not using the data from the backend properly.

jannisvisser:
@gulfaraz 
- the initial reasoning was that this 'trigger-dependent active=false to active=true setting' was not properly mocked by the mock-data. 
- You apparently also weren't able to solve that in your PR. 
- So I went for this as a solution, as updating the 'active' property in the front-end instead of in the back-end immediately solves this for mock-data as well ..
- .. while I do not see any downsides. But please enlighten me if you do.

gulfaraz:
@jannisvisser I see a downside in the design of this behaviour.

The database has an `active` property which is read by the API-service and passed to the IBF-dashboard. The IBF-dashboard should **not undermine** the value sent by the API-service.

One solution is to change the design such that the `active` property does not exist in the API-service or the database.
This would be a poor design choice as such control should not be given to the IBF-dashboard. A decision like "if active trigger then show a layer by default" is a behaviour not based on user interaction but a desired trait of the IBF-system which should be reflected in the API-service.

This reasoning also applies to the aggregate calculation we currently do on the IBF-dashboard, we should migrate such responsibilities to the API-service or a separate backend that is responsible for the _business logic_. The IBF-dashboard is only responsible for presenting the data provided by the API-service. An interface should not parse or mutate the data which comes from the _source of truth_.

jannisvisser:
@gulfaraz OK, then the conclusion is to close this PR.

PR #635, opened 2021-03-26
fix: refactor static data AB#6363

jannisvisser:


jannisvisser:
@gulfaraz In this case I thought it better to merge this PR already, because it came with so many data-migration changes that I wanted to make sure the test-vm is working again afterwards.
- Note btw, that to get your local db working again, I updated this section in the readme: https://github.com/rodekruis/IBF-system#note-on-local-database
- Still not 100% ideal, but already a lot more manageable then working with dump/restore..
- The next step would be to take out the runSetup.py step completely (only the EGY-trigger-level file is left to migrate)

PR #669, opened 2021-04-12
Fix.refactor static 2

jannisvisser:


services/API-service/src/scripts/seed-init.ts
```diff
@@ -179,11 +180,16 @@ export class SeedInit implements InterfaceScript {
...
     await seedRedcrossBranches.run();
 
-    // ***** SEED INDICATOR DATA PER ADMIN AREa *****
+    // ***** SEED INDICATOR DATA PER ADMIN AREA *****
     console.log('Seed Indicator data per admin-area...');
     const seedAdminAreaData = new SeedAdminAreaData(this.connection);
     await seedAdminAreaData.run();
 
+    // ***** SEED OTHER DATA *****
+    console.log('Seed other data...');
```
gulfaraz, line 189: _OTHER_ does not provide any useful information

Please specify what the data is.

services/API-service/src/api/rainfall-triggers/rainfall-triggers.entity.ts
```diff
@@ -0,0 +1,43 @@
+import { Entity, PrimaryGeneratedColumn, Column } from 'typeorm';
+
+@Entity('rainfallTriggers')
+export class RainfallTriggersEntity {
+  @PrimaryGeneratedColumn('uuid')
+  public id: string;
+
+  @Column()
```
gulfaraz, line 8: `countryCode` here accepts a random string

It should only accept values from the country table. [See example here](https://github.com/rodekruis/IBF-system/blob/7f70ea0389a97db0e5efa34afe31b2fc6d8ab3c2/services/API-service/src/api/admin-area-data/admin-area-data.entity.ts#L15-L17).

services/API-service/src/api/rainfall-triggers/rainfall-triggers.entity.ts
```diff
@@ -0,0 +1,43 @@
...
+  @Column({ type: 'real' })
+  public lat: string;
+
+  @Column({ type: 'real' })
+  public lon: string;
+
+  @Column()
+  public leadTime: string;
+
+  @Column({ nullable: true, type: 'real' })
```
gulfaraz, line 20: does a `null` `triggerLevel` makes sense in this context?

services/API-service/src/scripts/seed-other-data.ts
```diff
@@ -0,0 +1,56 @@
...
+              threshold99Perc: pixel['threshold_99perc'],
+              threshold2Year: pixel['2yr_threshold'],
+              threshold5Year: pixel['5yr_threshold'],
+              threshold10Year: pixel['10yr_threshold'],
+              threshold20Year: pixel['20yr_threshold'],
+              threshold50Year: pixel['50yr_threshold'],
+              threshold100Year: pixel['100yr_threshold'],
+            })
+            .execute()
+            .catch(console.error);
```
gulfaraz, line 49: catch `console.error` is not necessary any more because of [this](https://github.com/rodekruis/IBF-system/blob/7f70ea0389a97db0e5efa34afe31b2fc6d8ab3c2/services/API-service/ormconfig.js#L12)

services/IBF-pipeline/pipeline/lib/pipeline/dynamicDataDb.py
```diff
@@ -91,31 +91,13 @@ def processDynamicDataDb(self):
     def apiGetRequest(self, path, country_code):
         import requests
 
-        login_response = requests.post('http://12.0.0.8:3000/api/user/login', data=[('email',ADMIN_LOGIN),('password',ADMIN_PASSWORD)])
+        login_response = requests.post('http://ibf-api-service:3000/api/user/login', data=[('email',ADMIN_LOGIN),('password',ADMIN_PASSWORD)])
```
gulfaraz, line 94: The login url string should be configurable. Preferably as an ENV variable.

services/IBF-pipeline/pipeline/lib/pipeline/dynamicDataDb.py
```diff
@@ -91,31 +91,13 @@ def processDynamicDataDb(self):
...
         TOKEN = login_response.json()['user']['token']
 
-        response = requests.get('http://12.0.0.8:3000/api/'+path+'/'+country_code,headers={'Authorization': 'Bearer ' + TOKEN})
+        response = requests.get('http://ibf-api-service:3000/api/'+path+'/'+country_code,headers={'Authorization': 'Bearer ' + TOKEN})
```
gulfaraz, line 97: The api url string should be configurable. Preferably as an ENV variable.

services/API-service/src/api/rainfall-triggers/rainfall-triggers.entity.ts
jannisvisser: @gulfaraz but then how do I code this in the seed-script? I first need to get the full countryEntity, and pass that to the country-column in the rainfallTriggersEntity?

services/IBF-pipeline/pipeline/lib/pipeline/dynamicDataDb.py
jannisvisser: @gulfaraz I get that it's good to make it a configurable setting, but it's not server-dependent in this case right? Can it not go in settings.py?

services/API-service/src/api/rainfall-triggers/rainfall-triggers.entity.ts
gulfaraz: It should be possible to do this without fetching the `countryEntity`.
The definition in the `rainfallTriggersEnitity` is updated but the rest should work as is.

We're only using a _referenced string_ instead of _any string_ for `countryCode`.

jannisvisser:
@gulfaraz all your comments are processed (incl. replicating the same changes at other similar places)

jannisvisser:
@gulfaraz please merge only after the sprint review ;)

gulfaraz:
@jannisvisser do we still need data.zip for API-service? or does this PR completely remove our dependency on data.zip?

jannisvisser:
@gulfaraz it's indeed not necessary any more for database/API-service. But it is still needed of course for IBF-pipeline/geoserver.

services/API-service/src/scripts/seed-rainfall-data.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    // EGY
+    // Trigger levels per lat/lon pixel
+    const fileName = `./src/scripts/git-lfs/rainfall/rainfall_trigger_levels_EGY.csv`;
+    const data = await this.seedHelper.getCsvData(fileName);
+
+    await Promise.all(
+      data.map(
+        async (pixel): Promise<void> => {
+          return this.rainfallTriggersRepository
```
gulfaraz, line 31: @jannisvisser please try to use typeorm entities instead of query builder

services/API-service/src/scripts/seed-rainfall-data.ts
```diff
@@ -0,0 +1,55 @@
...
+            .createQueryBuilder()
+            .insert()
+            .values({
+              countryCode: 'EGY',
```
gulfaraz, line 35: @jannisvisser can we make this country agnostic? We could accept a `countryCodeISO3` parameter then we can support rainfall triggers for any country.

services/API-service/src/scripts/seed-init.ts
```diff
@@ -179,11 +180,16 @@ export class SeedInit implements InterfaceScript {
...
 
-    // ***** SEED INDICATOR DATA PER ADMIN AREa *****
+    // ***** SEED INDICATOR DATA PER ADMIN AREA *****
     console.log('Seed Indicator data per admin-area...');
     const seedAdminAreaData = new SeedAdminAreaData(this.connection);
     await seedAdminAreaData.run();
 
+    // ***** SEED RAINFALL DATA *****
+    console.log('Seed rainfall data...');
+    const seedOther = new SeedRainfallData(this.connection);
```
gulfaraz, line 190: `seedOther` to `seedRainfall`

services/IBF-pipeline/pipeline/lib/pipeline/rainfalldata.py
```diff
@@ -242,7 +240,7 @@ def findTrigger(self):
         # for leadtime in np.unique(df_thresholds.forecast_time.values):
 
         # threshold (1 degree)
-        df_leadtime = df_thresholds[df_thresholds.forecast_time == self.leadTimeValue]
+        df_leadtime = df_thresholds[df_thresholds['leadTime'] == self.leadTimeValue]
```
gulfaraz, line 243: @jannisvisser We should move these constant strings `leadTime` and `triggerLevel` to a separate constants file. Similar to how we have enums / models in TypeScript.

services/IBF-pipeline/pipeline/lib/pipeline/rainfalldata.py
```diff
@@ -261,21 +259,21 @@ def findTrigger(self):
                                 how='left', op='intersects')
 
         # interpolate NaN cells
-        known = ~np.isnan(compare_gdf['forecast_time'])
+        known = ~np.isnan(compare_gdf['leadTime'])
```
gulfaraz, line 262: @jannisvisser We should move these constant strings `leadTime` and `triggerLevel` to a separate constants file. Similar to how we have enums / models in TypeScript.

services/IBF-pipeline/pipeline/lib/pipeline/rainfalldata.py
```diff
@@ -261,21 +259,21 @@ def findTrigger(self):
...
         unknown = ~known
         z = griddata((compare_gdf['longitude'][known], compare_gdf['latitude'][known]),
-                     compare_gdf[self.TRIGGER_RP_COLNAME][known],
+                     compare_gdf['triggerLevel'][known],
```
gulfaraz, line 265: @jannisvisser We should move these constant strings `leadTime` and `triggerLevel` to a separate constants file. Similar to how we have enums / models in TypeScript.

services/IBF-pipeline/pipeline/lib/pipeline/rainfalldata.py
```diff
@@ -261,21 +259,21 @@ def findTrigger(self):
...
                      (compare_gdf['longitude'][unknown], compare_gdf['latitude'][unknown]))
-        compare_gdf[self.TRIGGER_RP_COLNAME][unknown] = z.tolist()
+        compare_gdf['triggerLevel'][unknown] = z.tolist()
```
gulfaraz, line 267: @jannisvisser We should move these constant strings `leadTime` and `triggerLevel` to a separate constants file. Similar to how we have enums / models in TypeScript.

services/IBF-pipeline/pipeline/lib/pipeline/rainfalldata.py
```diff
@@ -261,21 +259,21 @@ def findTrigger(self):
...
 
         compare_gdf[str(str(self.leadTimeLabel) + '_pred')] = np.where(
-            (compare_gdf['mean_by_day'] > compare_gdf[self.TRIGGER_RP_COLNAME]), 1, 0)
+            (compare_gdf['mean_by_day'] > compare_gdf['triggerLevel']), 1, 0)
```
gulfaraz, line 270: @jannisvisser We should move these constant strings `leadTime` and `triggerLevel` to a separate constants file. Similar to how we have enums / models in TypeScript.

services/IBF-pipeline/pipeline/lib/pipeline/rainfalldata.py
```diff
@@ -261,21 +259,21 @@ def findTrigger(self):
...
         compare_gdf['fc_day'] = compare_gdf['fc_day'].astype(str)
         df_trigger = compare_gdf.filter(
             ['latitude', 'longitude', 'geometry', str(str(self.leadTimeLabel) + '_pred')])
 
         # out = df_trigger.to_json()
-        # output_name = '%s_%sday_'%(runcycle_day, self.leadTimeLabel) +self.TRIGGER_RP_COLNAME
+        # output_name = '%s_%sday_'%(runcycle_day, self.leadTimeLabel) + 'triggerLevel'
```
gulfaraz, line 276: @jannisvisser Commented code can be removed. If needed in the future then we look for these in the git history.

gulfaraz (changes requested):
[The acceptance criteria according to the PBI](https://dev.azure.com/redcrossnl/IBF%20System/_workitems/edit/6363)
> IBF-pipeline can only connect to database via API-service

IBF-pipeline connects to the database [here](https://github.com/rodekruis/IBF-system/blob/fix.refactor-static-2/services/IBF-pipeline/pipeline/lib/notifications/getFloodInfo.py#L9), [here](https://github.com/rodekruis/IBF-system/blob/fix.refactor-static-2/services/IBF-pipeline/pipeline/lib/pipeline/dynamicDataDb.py) and [here](https://github.com/rodekruis/IBF-system/blob/fix.refactor-static-2/services/IBF-pipeline/pipeline/lib/setup/setupConnection.py)

> All database tables have a corresponding entity

There are no entities in the pipeline. API-service has some entities but still uses SQL queries.

> No more SQL queries

The IBF-pipeline and API-service still use SQL queries - [for example](https://github.com/rodekruis/IBF-system/blob/fix.refactor-static-2/services/API-service/src/api/data/data.service.ts).

jannisvisser:
@gulfaraz thanks for the review
- in general, for me it's always OK if you directly make changes
- in this particular case definitely, because I won't have much time soon, so then this item/PR would stay open unnecessarily long
- I agree with most of your comments in general, but do always think we have to strike the right balance between speed and quality.. I cannot assess completely now, whether processing all your ideas, will not slow us down too much, given also other priorities..
- I made a new subtask for replacing downloadGeoDataFromDb  by an endpoint in the next refactor item

gulfaraz:
@jannisvisser Great! If you don't mind, I'd like this bring this up with the team to find the right balance of speed and quality.

services/IBF-pipeline/pipeline/lib/pipeline/dynamicDataDb.py
```diff
@@ -91,31 +90,13 @@ def processDynamicDataDb(self):
...
-                             "\".\""+table+"\" LIMIT 0")
-            colnames = [desc[0] for desc in self.cur.description]
-            self.con.commit()
-            self.con.close()
-        except psycopg2.ProgrammingError as e:
-            logger.info(e)
-
-        return data, colnames
-
     def downloadGeoDataFromDb(self, schema, table, country_code=None):
```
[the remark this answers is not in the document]
jannisvisser: @gulfaraz I have moved the task into the follow-up refactor item

services/API-service/src/scripts/seed-init.ts
jannisvisser: done (when committed) > forgot this one in the previous editing round

services/IBF-pipeline/pipeline/lib/pipeline/rainfalldata.py
jannisvisser: done (when committed). I agree with your point of course. In this particular case though, I was adhering already a little bit to the idea that Phuoc is the owner of this code, and I changed only things that were needed to make things work. Anyway, it's done.

services/IBF-pipeline/pipeline/lib/pipeline/rainfalldata.py
jannisvisser: I have done this now for the examples you commented on (when committed). In general, I agree with the sentiment also. But maybe good to know that the context here is:
- the IBF-pipeline has quite low coding standards compared to other parts of our code. There's A LOT that could/should be improved, so it feels a bit random to start with such a small insignificant detail, and leave all the rest. But I guess we have to start somewhere..
- also Python (at least the way we use it?) is not strongly typed in any way, which makes using enums less relevant right? Because it will not give you errors when you mistype or something.

services/API-service/src/scripts/seed-rainfall-data.ts
jannisvisser: done (when committed). and I'm only seeding this data for the right countries, based on a new country-property which is at the moment 'glofas' or 'rainfall' (with its own enum ;) )

jannisvisser:
@gulfaraz I processed all the requested changes. can you review again?

the api-service check is failing because of this TypeORM nullable column quirk
- I added a column (hazardModel)
- which is not nullable
- but when TypeORM is adding the column (upon (re)start of the API-service application), while there are existing rows in the tables, it will find empty columns
- which is why the test fails
- I could solve this by using a default value, but found this weird.. (but maybe we should after all, Please do so if you see fit)

gulfaraz:
@jannisvisser took me a while to get around the database issues. Any idea how we'd get around those migration errors?

PR #674, opened 2021-04-13
fix: update URCS logo AB#7050

jannisvisser:


gulfaraz (approved):
The logo is updated as expected.

Screenshot 2021-04-13 at 15 13 00

Image URL for mail notification also works.

PR #675, opened 2021-04-13
fix: show indicator for stage environment AB#7065

gulfaraz:


jannisvisser:
@gulfaraz works locally (when I change the value around)
- is 'stage' the right value for this instead of 'staging'? (doesn't matter too much obviously)
- is this value already set as such on the staging-server?

jannisvisser (approved):

PR #685, opened 2021-04-19
fix: update EAP link UGA AB#7051

jannisvisser:


jannisvisser:
@gulfaraz we need to include this (very simple) change also still before deploy to staging & production

gulfaraz (approved):

PR #687, opened 2021-04-23
refactor: get glofas data from datalake AB#6512

jannisvisser:


gulfaraz:
`runPipeline.py` gives this error.

```
2021-04-26 12:24:27.482130
--------STARTING: ZMB--------------------------
--------STARTING: 3-day--------------------------
Download data failed. Trying again in 10 minutes.
<urlopen error ftp error: error_temp('451 File not found: ZambiaRedcross_glofas_point/glofas_pointdata_ZambiaRedcross_2021-04-2600.tar.gz',)>
```

Retrying for 12 hours for the above error doesn't make sense to me.
What are the expected outcomes when,
1. the file is present in the datalake
2. the file is not found in the datalake

jannisvisser:
@gulfaraz 
- first, your comment does not relate to a change in this PR (as this PR deals with the API which is not used for Zambia, which still uses the old FTP)
- the bug is because the FTP is unavailable a lot. This is really annoying, hopefully we can switch to API for Zambia soon also.
- and the 12 hours (which indeed is used both in the API/datalake case and in the FTP case) sort of makes sense. The idea is that the data delivery can be (very) late on each day, and we simply keep trying out for most of the day. What's wrong with that? 
- And to answer your specific question:
- 1. when the file is found it continues with the script
- 2. when the file is not found it will retry in 10 minutes

gulfaraz (approved):
works for UGA
```
2021-05-03 12:53:35.232237
--------STARTING: UGA--------------------------
--------STARTING: 5-day--------------------------
Downloading from datalake:  ibf/glofas/glofas-forecast-20210503.nc

Extracting Glofas Data

DWRM2
DWRM12
DWRM3
DWRM9
DWRM7
DWRM14
DWRM10
DWRM1
Extracted Glofas data - File saved
Extracted Glofas data - Trigger per day File saved
Processed Glofas data - File saved
Total flood extent file written
indicator:  population
triggers_rp_per_station uploaded
triggers_per_day uploaded
calculated_affected uploaded
SQL EXECUTED
75.49357771873474
```

PR #692, opened 2021-04-26
fix: remove ion-content from app component html AB#7219 AB#7220 AB#7227

gulfaraz:


jannisvisser:
@gulfaraz did this solve all 3 sub-bugs at once? either way, nice! Cannot check anything now though, because ibf-test is down.

PR #694, opened 2021-04-27
Era

RubenGeo:


docker-compose.yml
```diff
@@ -43,6 +43,8 @@ services:
             pipeline-network:
         depends_on:
             - ibf-geoserver
+        volumes:
+            - ./services/IBF-pipeline/pipeline/lib/pipeline:/home/ibf/pipeline
```
jannisvisser, line 47: named volume

jannisvisser:
@RubenGeo this latest commit, is the result after I added the store+layer myself manually in geoserver again. Not clear what the difference is, except for maybe a different coverageStore ID or something. 

Anyway, if this change also works for you (maybe delete the store + layer first before pulling this), then we can simply try and merge and see if it works on test-vm.

PR #707, opened 2021-05-03
fix: don't try download if mock AB#7447

jannisvisser:


jannisvisser:
@gulfaraz 
- this fix is meant to change the weird fact that even in 'mock' setting, the pipeline tries to download data first
- this thereby answers to Orla's request to have access to Zambia even though the FTP keeps failing now
- also, it gives us developers the chance to more easily surpass FTP/API issues
- I think it makes sense to approve & merge this before the next deploy to staging/production, so that Orla can keep having access to Zambia on staging.

gulfaraz (changes requested):
Screenshot 2021-05-03 at 16 56 09

jannisvisser:
@gulfaraz I corrected and tested it

gulfaraz (approved):
I did run into other issues but if I manually select the correct lead time I see the desired view.
screenshot-localhost_8080-2021 05 03-17_34_38

PR #708, opened 2021-05-04
build: missing yargs package

gulfaraz:


jannisvisser:
@gulfaraz do you have some explanation on the origin/purpose? And please let me know if a review is urgent, then I can maybe get to it before Friday. Otherwise I'll leave it.

gulfaraz:
@jannisvisser the yargs package is used [here](https://github.com/rodekruis/IBF-system/blob/master/services/API-service/src/scripts/scripts.module.ts#L3) and [here](https://github.com/rodekruis/IBF-system/blob/master/services/API-service/src/scripts.ts#L3)

on a clean setup, I got the error message `Cannot find module 'yargs'`

running `npm install --save-dev yargs` resolved this problem so I believe it needs to be a part of `package.json`

jannisvisser (approved):
Successfully rebuilt + started IBF-API-service.

PR #719, opened 2021-05-10
Test.cypress setup

jannisvisser:
@gulfaraz I thought I would align the front-end port number to use to be always 4200, instead of sometimes 8080 and sometimes 4200. 

Additionally I updated the readme. Feel free to add more if you think it's relevant.


gulfaraz (approved):
I did not need to run Step 0 for my setup but I imagine it cannot harm to have it as Step 0.

Does the port number change affect any of the environment variables in the deployments? If yes, please update them. If no, we're good to merge.

PR #720, opened 2021-05-10
fix: use rxjs of instead of EMPTY AB#7227

gulfaraz:
@jannisvisser I had used EMPTY observable which doesn't trigger the update layer function if moving from active to inactive state for the first time. This is why this behaviour only occurred on the population exposed layer and in the trigger state (as it is enabled by default).

Using `of(null)` resolves this issue. Please review and confirm if this can be considered fixed.


jannisvisser:
@gulfaraz unfortunately this doesn't seem to change anything on my end (with the local setup connected to test-db). Should we call after to identify the difference?

jannisvisser (changes requested):
as discussed: legend is still visible after 1st toggle-off, but disappears with 2nd toggle-off

gulfaraz:
@jannisvisser The legend update code requires data to not be null. Please review again.

PR #747, opened 2021-06-07
Refactor.migration scripts

jannisvisser:


jannisvisser:
I have also deleted all DB_* secrets from Github > settings > secrets.

gulfaraz (commented):
@jannisvisser after running seed to initialize the database. I was able to run the existing migration scripts.

I also tested generating the migration scripts after adding a new column to an entity. Works as expected.

PR #748, opened 2021-06-08
Feat.wording changes

jannisvisser:


jannisvisser:
This PR builds on (and thus includes also the changes from) #747 . Look at that one first. These changes needed a datamodel change, so I thought it would be good to immediately implement/test the migration-approach, which I did here. 

In [this commit](https://github.com/rodekruis/IBF-system/pull/748/commits/65be15f4bd3bbe3b6bc3ac089417347897124208) I removed unnecessary lines from the migration-script, which are added in every migration-script again, simply because TypeORM is somehow making these changes after initializing the tables.

services/API-service/src/api/country/country.entity.ts
```diff
@@ -38,11 +38,10 @@ export class CountryEntity {
...
   public defaultAdminLevel: AdminLevel;
 
-  @Column('text', {
-    array: true,
-    default: (): string => 'array[]::text[]',
+  @Column('json', {
+    default: {},
   })
-  public adminRegionLabels: string[];
+  public adminRegionLabels: JSON;
```
gulfaraz, line 44: Okay for now. When possible, please prefer to define a type object in the spirit of typescript.

gulfaraz (commented):
Screenshot 2021-06-11 at 12 25 43
Screenshot 2021-06-11 at 12 27 28
Wording interpolation works as expected.

PR #742, opened 2021-06-04
Feat.wards zambia

jannisvisser:

PR #749, opened 2021-06-10
Refactor.admin levels

gulfaraz:
@jannisvisser this PR contains part of the #742 review changes


jannisvisser (approved):
See comments in other PR

services/API-service/src/api/admin-area/admin-area.service.ts
```diff
@@ -132,32 +132,40 @@ export class AdminAreaService {
...
+        leadTime,
+      );
+
+      if (placeCodes.length) {
+        adminAreasScript = adminAreasScript.andWhere(
+          'area."placeCode" IN (:...placeCodes)',
+          { placeCodes: placeCodes },
+        );
+      }
+    } else {
```
jannisvisser, line 166: @gulfaraz It's especially this else-clause that I just meant with the hacky solution. It's a solution to errors that would otherwise appear in the front-end (about 'population_affected' missing.) It would be better to solve them there actually (I guess something with splitting the mapService.setAdminRegionStyle method between boundary layers and aggregates layers). But I did not have time to look at this properly.

gulfaraz: @jannisvisser I agree. Your commits are sufficient given our requirements. For a proper solution, this would need to be discussed in a new PBI/spike. Please feel free to create one if you want to prioritize it.

PR #751, opened 2021-06-10
refactor: all admin regions are equal

gulfaraz:
@jannisvisser This PR contains changes related to the review on #742


gulfaraz:
There are a few subscriber related changes that could help but then the refactor becomes too complex to review. So I've limited this PR to these changes. Any more refactoring can be done on a new PBI. Unless you disagree then please let me know what to prioritize.

jannisvisser:
@gulfaraz commenting here on both PRs, because I only looked at them together
- I see that you changed the functionality such that only 1 adminBoundary is toggled on at the same time. I get the point, but it doesn't all feel super-intuitive yet. (How this is unrelated from the fill-layers which are per definition on adminLevel 2). I would advise to explicitly ask for feedback on this in the Sprint Review.
- One specific downside of your new approach where all adminLevels are the same, is that the admin-area-popup for districts now only shows the name (if districts is toggled on). To also see the selected indicator (like before) you need to untoggled the district-boundary layer. (see 2 screenshots). While I agree with your general appraoch, I think this is not an ideal consequence, can we solve this in any way?
image
image
- Lastly, I found out that the data-side all of this is actually working correctly in part because of a major quirk in code that needs to be solved. I'll create + describe in a separate task.

jannisvisser (approved):

PR #764, opened 2021-06-21
Dengue > feedback from jannis and user test

RubenGeo:


services/API-service/src/api/admin-area/admin-area.service.ts
```diff
@@ -107,7 +107,6 @@ export class AdminAreaService {
       .where('area."countryCodeISO3" = :countryCodeISO3', {
         countryCodeISO3: countryCodeISO3,
       })
-      .andWhere('date = current_date')
```
jannisvisser, line 110: @RubenGeo You cannot just omit this where-clause. I understand what you want, but what you need here is to filter on the most recent date then instead of the current_date. Because otherwise it will return multiple rows for multiple dates instead of just the one you need.

PR #770, opened 2021-06-25
fix: drop table if exists

gulfaraz:
This change allows running seed on a non-empty database.


jannisvisser:
@gulfaraz I came accross this recently too, but it's not as simple as that.. 
- Because `const entities = this.connection.entityMetadatas` is already only supposed to retrieve entities that do exist, 
- so the `IF EXISTS` should not matter.
- I've ran the seed-script many of times while it was not empty, and successfully
- But yet I admit, that I came across this bug recently too, but then later it also worked again. I didn't find out the pattern.
- I guess you could say though that your proposal will never hurt. So in that case I approve.

jannisvisser (approved):

PR #771, opened 2021-06-28
Feat.population total

gulfaraz:
Total population should be visible in the indicator/aggregates column, the matrix component and the map with legend.

screenshot-localhost_4200-2021 06 28-12_33_52


jannisvisser:
@gulfaraz without having looked any further, in your screenshot I don't see the 'Population' (grid) layer any more?

gulfaraz:
@jannisvisser Right, I probably used the same indicator names. Thank you.

gulfaraz:
@jannisvisser does it make sense to have two layers for the total population? If the information is redundant, should we prefer to remove the raster layer?

gulfaraz:
screenshot-localhost_4200-2021 06 28-17_33_35

jannisvisser:
Some comments
- I see that for Zambia there is also populationTotal data uploaded for provinces and wards, but all under "adminLevel"=2, which is incorrect. I imagine this won't matter in the dashboard, because it takes only the right placeCodes, but should be fixed.
- The label is '% of Exposed population', while it should by '% of total population' 
- when I checkout the branch, run seed, and the trigger mock endpoint, it gives me 0% everywhere (for Zambia again). I see you screenshot, so it must be something wrong in my environment. I don't have more time now though to investigate.
- I would maybe go for 1 decimal in the percentage. If generally all percentages are around 0, 1, 2 or 3 % that gives a bit more differentation (and especially when it's rounded off to 0, that confuses people I think).

jannisvisser:
The 0% is probably related to this: the layer is not loading at all in the map.
image

gulfaraz:
@jannisvisser Please review when possible. Thank you!

> I see that for Zambia there is also populationTotal data uploaded for provinces and wards, but all under "adminLevel"=2, which is incorrect. I imagine this won't matter in the dashboard, because it takes only the right placeCodes, but should be fixed.

I've updated the Population seed data with the correct `AdminLevel` values.

> The label is '% of Exposed population', while it should by '% of total population'

Fixed

> when I checkout the branch, run seed, and the trigger mock endpoint, it gives me 0% everywhere (for Zambia again). I see you screenshot, so it must be something wrong in my environment. I don't have more time now though to investigate.

I was unable to reproduce this. I agree that this is probably unrelated. Makes sense to create a bug to reproduce this.

> I would maybe go for 1 decimal in the percentage. If generally all percentages are around 0, 1, 2 or 3 % that gives a bit more differentation (and especially when it's rounded off to 0, that confuses people I think).

I agree. It makes sense to make this change a separate item. Or as a part of [AB#7172](https://dev.azure.com/redcrossnl/IBF%20System/_workitems/edit/7172/) and [AB#7173](https://dev.azure.com/redcrossnl/IBF%20System/_workitems/edit/7173/)

jannisvisser:
@gulfaraz see here the changes to the mock-endpoint
- It's not very nice, because you basically have to duplicate this work in both pipeline and mock-endpoint at the moment
- But good enough for now. 
- Please note, that the changes around PHL I made in the mock-endpoint, were to clean it up a little bit. I find it quite messy how Ruben coded this. Specifically: the 'upload-exposure-PHL-potential-cases-threshold.json' file is not actually mock-data, but is actually trigger-levels per admin-area, which should be in seed-data area. It's good to realize this because soon we'll have to copy this behaviour for Ethiopia ERA..

jannisvisser (approved):
if you agree with my changes again, then you can merge

jannisvisser:
@gulfaraz sorry, I had to merge this, because I have to touch the same code again now for other reasons (the mock endpoint for multi-hazard). Which would have been extremely unhandy to do otherwise.

PR #780, opened 2021-07-02
Feat/8184 - Added links for Video and PDF download in email-template

SwatiTh:


jannisvisser (changes requested):
Technically it all works and looks good. Nice!
image

In terms of design/wording, some comments from me. But please get in touch with Priyanka indeed
- Both PDF/Video are instructions. So 'View flood status by watching video' is not correct. More something like 'video instructions for using dashboard'
- same for PDF. Wording should be more descriptive then 'donload pdf from here'.
- I'm exactly not sure if these 2 links should have their own button in this list. Maybe they can just be added to the description of 'View dashboard'. So something like 'See locations potentiall exposed in dashboard (see instructions in **video** and in **pdf**)', where the bold parts are then links.
- But again, please discuss with Priyanka, but do take my comments into that discussion.

jannisvisser (approved):

PR #781, opened 2021-07-05
feat #8181 uga population for new boundaries

boukepieter:


gulfaraz (changes requested):
@boukepieter the admin level column is missing in the new population file. Please add the columns for admin level and its values.

boukepieter:
@gulfaraz it wasn't in the old one either. It can be derived from the placeCode column, right? Or how do you want it otherwise? Just a column called adminLevel with a number between 1-4?

gulfaraz:
@boukepieter it is in the old one - [please see here](https://github.com/rodekruis/IBF-system/blob/master/services/API-service/src/scripts/git-lfs/admin-area-data/population_UGA.csv)

boukepieter:
@gulfaraz ah interesting, thanks. Wasn't aware of that, will change it accordingly.

PR #787, opened 2021-07-06
fix: use migrations properly with init AB#8485

gulfaraz:


jannisvisser:
@gulfaraz did you intentially remove the cleanAll step as well? This way, if you run the seed-script on a non-empty database it will fail, as it's simply adding the same data a 2nd time, which leads for example to uniqueness constraints.

I often use the seed-script to "reset" this way, not only to "set" on an empty database, so I prefer to keep the cleanAll step. 

PS: you can remove seedHelper from seed-init.ts

gulfaraz:
> did you intentionally remove the cleanAll step as well? This way, if you run the seed-script on a non-empty database it will fail, as it's simply adding the same data a 2nd time, which leads for example to uniqueness constraints.

Yes, it made sense to prefer the risk of failing on seed rather than drop the database.

I often use the seed-script to "reset" this way, not only to "set" on an empty database, so I prefer to keep the cleanAll step.

> I can add it back if you prefer it. Let me know. @jannisvisser

jannisvisser:
@gulfaraz Do you have a quick alternative to reset then? Because the quickest I would know now is to drop + recreate the schema and then run seed. I would vote to put it back.

gulfaraz:
@jannisvisser I do the same (drop and create). I'll put the reset back.

jannisvisser (approved):

PR #788, opened 2021-07-06
feat: add admin boundaries and enable admin levels for uga AB#8504

gulfaraz:
@jannisvisser Enabled Admin Levels 1, 3 and 4 for Uganda.

uga_adm1
uga_adm3
uga_adm4


jannisvisser (changes requested):
@gulfaraz this looks good for the most part, but there's an important thing missing.
- because the pcodes of admin level 2 have changed
- this means that all data that is joined to that on pcode must change accordingly as well
- this is done now only for the glofas station to district mapping
- and for total population Bouke made another PR (not sure what the status of that was)
- but for other data (vulnerability and its underlying indicators + covid I think) we need to make the transition using oldADM2_PCODE that Bouke provided I think

jannisvisser:
I'm merging, as this is reviewed by Swati

PR #789, opened 2021-07-09
fix: truncate instead of drob before seed AB#8485

jannisvisser:


jannisvisser:
@gulfaraz needed to merge this to fix test-vm

gulfaraz:
@jannisvisser This change does not make sense to me. The point of `DROP` was to make sure that the tables don't already exist when applying the init migration. With `TRUNCATE` only empties the existing tables without removing them. So running the init migration on a truncated database will fail.

jannisvisser:
@gulfaraz without this change, the seed-script was simply not working because
- it first dropped all tables
- but then did not recreate them anymore like before, because we removed the synchronize-step in seed-init

gulfaraz:
The seed script is not responsible for creating the tables. That is the responsibility of the migration script. @jannisvisser

PR #795, opened 2021-07-13
feat: use ftp instead of datalake AB#8575

gulfaraz:


jannisvisser:
@gulfaraz looks good (assuming you've tested that the pipeline works on non-mock for a flood-country). I'm realizing now, this should actually come with a (hotfix) deploy to production, otherwise there was no point in doing this quickly. We should then cherrypick this commit and merge in the current production-version, because otherwise we are stuck with the scenario we discussed yesterday where we would have to deploy the half-baked multi-hazard solution...

PS: see latest status of the logic app discussion [here](https://teams.microsoft.com/l/message/19:072bae37d785435ba5c3d9d6de83e50b@thread.skype/1626169250643?tenantId=d3ab9790-6ae2-4bd8-aa5e-02864483e7c7&groupId=1e24e23a-6105-4743-82ec-c9c9383b0750&parentMessageId=1626086255583&teamName=510%20-%20Team&channelName=%5BADM%5D%20Technical%20Infrastructure&createdTime=1626169250643). 
- It is actually solved now
- except for a small question around publishing the needed change in the ADF-pipeline ..
- it that would be resolved (and the change published by Maarten or by myself if it happens today/tomorrow)
- then the whole problem is actually resolved, and we don't need to deploy to production and we can switch this back again ...

gulfaraz:
@jannisvisser That's nice, please let me know if the change gets pushed or not. I'm following the discussion anyway. I'll hold the hotfix deployment until tomorrow. Thank you for the update.

I've tested the pipeline on real data for Uganda and Zambia.

PR #810, opened 2021-07-20
Feature/sthakre/8470

SwatiTh:
Improve hazard type switch - impl


interfaces/IBF-dashboard/src/app/config.ts
```diff
@@ -43,3 +43,10 @@ export const LEAFLET_MAP_OPTIONS: MapOptions = {
   zoom: 5,
   layers: [],
 };
+
+export const DISASTER_TYPES_SVG_MAP = {
+  'floods': '/assets/icons/Flood.svg',
+  'dengue': '/assets/icons/Mosquito.svg',
+  'malaria': '/assets/icons/Mosquito.svg',
+  'heavy rainfal': '/assets/icons/Alert.svg',
```
gulfaraz, line 51: Please use the [Disaster Type Key](https://github.com/rodekruis/IBF-system/blob/master/interfaces/IBF-dashboard/src/app/types/disaster-type-key.ts#L1) instead of the Disaster Type Label.

This will help keep consistent and also avoid typos as in [line 51](https://github.com/rodekruis/IBF-system/pull/810/files#diff-f64cdb1ffcfa51a8e45821b7790d549ca8f6f3b2aff5b34164710ef45ccea56eR51).

interfaces/IBF-dashboard/src/app/components/disaster-type/disaster-type.component.ts
```diff
@@ -37,5 +40,6 @@ export class DisasterTypeComponent implements OnInit, OnDestroy {
 
   public switchDisasterType(disasterType: DisasterType): void {
     this.disasterTypeService.setDisasterType(disasterType);
+    this.selectedDisasterType = disasterType.label;
```
gulfaraz, line 43: [Please use Disaster Type Key instead of Disaster Type Label here.](https://github.com/rodekruis/IBF-system/blob/master/interfaces/IBF-dashboard/src/app/models/country.model.ts#L44)

interfaces/IBF-dashboard/src/app/components/disaster-type/disaster-type.component.ts
```diff
@@ -11,6 +12,8 @@ import { CountryService } from '../../services/country.service';
 })
 export class DisasterTypeComponent implements OnInit, OnDestroy {
   public disasterTypes: DisasterType[] = [];
+  public disasterTypeMap = DISASTER_TYPES_SVG_MAP
+  public selectedDisasterType: string = null;
```
gulfaraz, line 16: I assumed that there is always a default disaster type based on the selected country. A situation where there is no selected disaster type probably never occurs.

Please use [Disaster Type](https://github.com/rodekruis/IBF-system/blob/master/interfaces/IBF-dashboard/src/app/models/country.model.ts#L43) as the type for `selectedDisasterType`.

interfaces/IBF-dashboard/src/app/components/disaster-type/disaster-type.component.scss
```diff
@@ -0,0 +1,3 @@
+.active {
+    background-color: rgb(244, 231, 254);
```
gulfaraz, line 2: Please only use the colours from [variables.scss](https://github.com/rodekruis/IBF-system/blob/master/interfaces/IBF-dashboard/src/theme/variables.scss). [Refer to the Ionic documentation on how to use colours.](https://ionicframework.com/docs/theming/colors)

interfaces/IBF-dashboard/src/app/components/disaster-type/disaster-type.component.html
```diff
@@ -1,6 +1,10 @@
-<ion-tab-bar *ngFor="let disasterType of disasterTypes">
-  <ion-tab-button (click)="switchDisasterType(disasterType)">
-    <ion-icon name="warning"></ion-icon>
-    <ion-label>{{ disasterType.label }}</ion-label>
+
+<ion-tab-bar *ngFor="let disasterType of disasterTypes"
+  [ngClass]="{'active': disasterType.label === selectedDisasterType}">
```
gulfaraz, line 3: You can ignore this comment if you choose to. I prefer to use a [consistent class conditional syntax format](https://stackoverflow.com/a/41974490/1753041). [Please take a look at how we've used class conditionals in our other components for reference.](https://github.com/rodekruis/IBF-system/blob/master/interfaces/IBF-dashboard/src/app/components/timeline/timeline.component.html#L18)

interfaces/IBF-dashboard/src/app/components/disaster-type/disaster-type.component.html
```diff
@@ -1,6 +1,10 @@
...
+  <ion-tab-button (click)="switchDisasterType(disasterType)" [ngClass]="{'active': disasterType.label === selectedDisasterType}">
```
gulfaraz, line 4: Is this conditional class an accidental duplicate? I don't see the need for two conditional classes which make the same check.

gulfaraz (changes requested):
@SwatiTh Please let me know if any of my review comments is unclear. Most of the changes look good enough to merge.

Screenshot 2021-07-26 at 11 39 52


I noticed that the end layout is different from the figma layout. Is the decision to ignore the figma decision intentional? If so, why?

interfaces/IBF-dashboard/src/app/config.ts
SwatiTh: Taken care

interfaces/IBF-dashboard/src/app/components/disaster-type/disaster-type.component.ts
SwatiTh: Taken care

interfaces/IBF-dashboard/src/app/components/disaster-type/disaster-type.component.html
SwatiTh: Did it intentionally.

interfaces/IBF-dashboard/src/app/components/disaster-type/disaster-type.component.scss
SwatiTh: Taken care

interfaces/IBF-dashboard/src/app/components/disaster-type/disaster-type.component.ts
SwatiTh: Taken care

jannisvisser:
@SwatiTh I've checked this out locally, but it seems the disaster-tabs are still below each other, while I believe the figma-file @gulfaraz was referring to earlier, has them side-by-side? Can you link me to the figma-file you were provided?

jannisvisser (approved):
@gulfaraz @SwatiTh accepting this, as I understood from Swati that Stefania was fine with this temp-solution and there are new items created for updating this.

PR #816, opened 2021-07-27
test: lead time component integration tests

SUSAGRAW:
Added : Video Guide also


jannisvisser:
@SUSAGRAW 
- this PR is named after the lead-time component, but also includes other things (total population) / (video guide). Can you please organize this better?
- specifically leadTime I only see very little tested. You test if there are timline buttons visible at all. And then you click on each button, but it's unclear what you want to test what should happen upon the click?

PR #823, opened 2021-08-02
fix: get latest update of threshold value instead of first AB#8876

jannisvisser:


jannisvisser:
I'm merging this immediately, so I can test on test-vm whether this fix solved the issue, as I couldn't reproduce this locally, only on server.

PR #838, opened 2021-08-12
feat: add Zimbabwe + droughts AB#7438

jannisvisser:


services/IBF-pipeline/geoserver-workspaces/ibf-system/population_ZWE/population_ZWE/coverage.xml
```diff
@@ -0,0 +1,119 @@
+<coverage>
```
SwatiTh, line 1: @jannisvisser, Do we need this coverage file as part of the PR? Is there any specific reason?

SwatiTh (approved):
LGTM

services/IBF-pipeline/geoserver-workspaces/ibf-system/population_ZWE/population_ZWE/coverage.xml
jannisvisser: @SwatiTh these files in /geoserver-workspaces/ are created automatically after adding a layer manually in geoserver (on localhost:8081/geoserver). By committing them, you don't have to repeat that manual process on every server.

PR #853, opened 2021-08-20
docs: to help while installation, guidline provided in troubleshoot file

SwatiTh:
Please review

### DIFF
--- a/docs/TROUBLESHOOT.md
+++ b/docs/TROUBLESHOOT.md
@@ -1,0 +1,74 @@
+### Description
+
+Upon running the <docker-compose -f docker-compose.yml -f docker-compose.override.yml up> all the containers are not started mentioned in the docker-compose file. Some containers throw certain errors mentioned below
+
+### To Start/Run the app
+
+◦	docker-compose up -d
+◦	docker-compose stop ibf-dashboard
+◦	cd interfaces/IBF-dashboard
+◦	npm start
+
+### Expected behaviour
+All containers mentioned in the docker-compose file /IBF-system/docker-compose.yml be up and running.
+Application should load and respond properly on the localhost:4200.
+
+### Desktop (please complete the following information):
+
+OS: MAC OS
+Browser : chrome
+Version : 92.*
+
+### Workaround
+    ◦	If the containers ibf-system_ibf-pipeline , ibf-system_ibf-api-service do not start.
+    ◦	If the container do not start automatically via the 'docker compose up' command. The workaround for this would be, manually start the container and run this command.
+
+ docker-compose -f docker-compose.yml -f docker-compose.override.yml up
+
+### After step2 of readme if there is any problem to run the app 
+You can try below steps:
+    ◦	throw away "IBF-app" schema of your local database : drop schema "IBF-app" cascade;
+    ◦	recreate "IBF-app" schema: create schema "IBF-app";
+    ◦	select * from "IBF-app".disaster;
+	◦	restart api-service: "docker-compose restart ibf-api-service"
+    ◦	run: docker-compose exec ibf-api-service
+	◦	wait until done, check with: "docker-compose logs -f ibf-api-service"
+	◦	run seed script: "docker-compose exec ibf-api-service npm run seed"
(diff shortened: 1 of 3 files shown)


docs/TROUBLESHOOT.md
```diff
@@ -0,0 +1,74 @@
+### Description
+
+Upon running the <docker-compose -f docker-compose.yml -f docker-compose.override.yml up> all the containers are not started mentioned in the docker-compose file. Some containers throw certain errors mentioned below
+
+### To Start/Run the app
+
+◦	docker-compose up -d
+◦	docker-compose stop ibf-dashboard
```
jannisvisser, line 8: Please add something here, that the bottom 3 lines of these 4 are just a suggestion, with the benefit that changes in front-end code are immediately reflected.

docs/TROUBLESHOOT.md
```diff
@@ -0,0 +1,74 @@
...
+    ◦	If the containers ibf-system_ibf-pipeline , ibf-system_ibf-api-service do not start.
+    ◦	If the container do not start automatically via the 'docker compose up' command. The workaround for this would be, manually start the container and run this command.
+
+ docker-compose -f docker-compose.yml -f docker-compose.override.yml up
+
+### After step2 of readme if there is any problem to run the app 
+You can try below steps:
+    ◦	throw away "IBF-app" schema of your local database : drop schema "IBF-app" cascade;
+    ◦	recreate "IBF-app" schema: create schema "IBF-app";
+    ◦	select * from "IBF-app".disaster;
```
jannisvisser, line 32: what is this? just to check if the table exists or something? If so, add that in. Besides, this table cannot exist right after creating the schema. There must be a step in between where the ibf-api-service is restarted (so that the migration-scripts are run). Please add that in.

docs/TROUBLESHOOT.md
```diff
@@ -0,0 +1,74 @@
...
+	◦	restart api-service: "docker-compose restart ibf-api-service"
+    ◦	run: docker-compose exec ibf-api-service
```
jannisvisser, line 34: this is not a possible command

docs/TROUBLESHOOT.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Workaround
+    ◦	If the containers ibf-system_ibf-pipeline , ibf-system_ibf-api-service do not start.
+    ◦	If the container do not start automatically via the 'docker compose up' command. The workaround for this would be, manually start the container and run this command.
+
+ docker-compose -f docker-compose.yml -f docker-compose.override.yml up
+
+### After step2 of readme if there is any problem to run the app 
+You can try below steps:
+    ◦	throw away "IBF-app" schema of your local database : drop schema "IBF-app" cascade;
```
jannisvisser, line 30: Recreating schema would only be a solution if there's something wrong with your existing setup. So this would never happen on initial first-time set up I guess.. Anyway, fine to leave this in here I guess.

docs/TROUBLESHOOT.md
```diff
@@ -0,0 +1,74 @@
...
+    ◦	recreate "IBF-app" schema: create schema "IBF-app";
+    ◦	select * from "IBF-app".disaster;
+	◦	restart api-service: "docker-compose restart ibf-api-service"
+    ◦	run: docker-compose exec ibf-api-service
+	◦	wait until done, check with: "docker-compose logs -f ibf-api-service"
+	◦	run seed script: "docker-compose exec ibf-api-service npm run seed"
+	◦	run mock-endpoint : http://localhost:3000/docs/#/scripts/ScriptsController_mockDynamic
```
jannisvisser, line 37: You could give a bit more explanation on how to run the endpoint maybe?

Or: maybe that information should go in the main readme? As part of the set-up steps. something like:
- set up system (docker-compose up)
- seed basic / static data (run seed)
- load 1st set of dynamic data to get a working dashboard (> run pipeline or mock-endpoint, > do this per country, per disasterType)

jannisvisser (changes requested):

jannisvisser (approved):

PR #864, opened 2021-09-02
feat: 8281 - Implemented Dam Layer  

SwatiTh:


jannisvisser (changes requested):
@SwatiTh very nice! some small changes needed though.
- the layer is active by default now, it should not be
- the info popup text is not correct, it seems to be copied from the 'drought_vulnerability_index'. Please get the right description from Marijke and in the meantime just use something minimal but true (e.g. "Dam locations")
- in the dam-popup you also list dynamic properties (Current capacity / Percentage full) even though this item is only supposed to be about the static part. The design for the dynamic part is still coming up, so better to not confuse people by already doing some of it. In fact, it is also better to remove the dynamic parts completely also from the csv, the dam.entity.ts, etc. 
- Don't include the countryCode ZWE in that popup, it adds no value.
- The 'dam name' is now both in the title and as a separate property. That is duplicate, remove the property.
- You can add the 'full capacity' but please add the right unit behind it (is it cube meters? so m3?)

PR #871, opened 2021-09-13
fix: show triggered disaster and fix left to right order of disaster …

SwatiTh:
…[AB#9317](https://dev.azure.com/redcrossnl/056030a3-96d2-4070-8718-48f602a2af75/_workitems/edit/9317) [AB#9318](https://dev.azure.com/redcrossnl/056030a3-96d2-4070-8718-48f602a2af75/_workitems/edit/9318)

### DIFF
--- a/interfaces/IBF-dashboard/src/app/components/disaster-type/disaster-type.component.ts
+++ b/interfaces/IBF-dashboard/src/app/components/disaster-type/disaster-type.component.ts
@@ -28,7 +28,7 @@ export class DisasterTypeComponent implements OnInit, OnDestroy {
   ngOnInit() {
     this.countrySubscription = this.countryService
       .getCountrySubscription()
-      .subscribe(this.onCountryChange); 
+      .subscribe(this.onCountryChange);
   }
 
   ngOnDestroy() {
@@ -37,7 +37,7 @@ export class DisasterTypeComponent implements OnInit, OnDestroy {
 
   private onCountryChange = (country: Country) => {
     if (country) {
-      this.disasterTypes = country.disasterTypes
+      this.disasterTypes = country.disasterTypes;
       this.disasterTypes.forEach((disasterType) => {
         this.eventService.getTriggerByDisasterType(
           country.countryCodeISO3,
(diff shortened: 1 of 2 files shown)


jannisvisser (changes requested):
@SwatiTh the 'show triggered disasterType by default' part works, however:
- I think there was miscommunication on the other requirement. Because the point was that floods should always be on the left, malaria on the right (of course not hard-coded like that, but based on the order they are entered in countries.json)

I also have some comments on the way this is coded
- You specifically use 'floods' in your code, but this is not robust. This code should also work in the future for a case where there are multiple disasterTypes of which floods is not one.
- I'm not sure that what extent using a timeout is best practice. There should always be a more robust solution, which actually knows that the required input is ready, instead of 'hoping' it's ready after 2000ms. @gulfaraz do you have input on this point?
- Please split your code in smaller functions (with clear names), instead of writing one long function. At least the whole new part you added should be split off in a separate function, but maybe even in multiple. This helps the readability.

jannisvisser (changes requested):
@SwatiTh this works now, except that when both are non-triggered I see neither as the selected disaster.
image

jannisvisser (changes requested):
@SwatiTh now it's wrong again in the following screenshot. Malaria is not selected while it is the only one triggered.

image

SwatiTh:
Hi @jannisvisser ,  Please check again, I can see the selection.
malaria-triggered

interfaces/IBF-dashboard/src/app/components/disaster-type/disaster-type.component.ts
```diff
@@ -35,16 +36,32 @@ export class DisasterTypeComponent implements OnInit, OnDestroy {
...
+        this.selectedDisasterType = this.disasterTypes[0].disasterType;
+      }
+    }
+    this.disasterTypesCounter++;
+  };
+
   private onCountryChange = (country: Country) => {
     if (country) {
       this.disasterTypes = country.disasterTypes;
       this.disasterTypes.forEach((disasterType) => {
```
jannisvisser, line 58: @SwatiTh Maybe you should before this line order the disasterTypes (alphabetically) to make sure they're always in the same order.

jannisvisser (approved):
@SwatiTh looks good now!

Not sure if this latest commit also fixed the default selection issue (not sure how), but either way it works now. Go ahead and merge.

image

PR #874, opened 2021-09-14
feat: reuse popover for consistent style AB#9327

gulfaraz:


SwatiTh (commented):
Looks good to me :)
middle-column-popover
Layer-popover

PR #882, opened 2021-09-21
fix: input autofill style should be readable AB#9486

gulfaraz:


SwatiTh:
Looks Good To Me
[
Login-id-password-colour-changes
](url)

PR #884, opened 2021-09-22
Add-EAP-action-6134, Change-trigger-msg-6380, Change-title-to-bold-9117

SwatiTh:


SwatiTh:
@jannisvisser @gulfaraz - seems like the CI checks while running npm test are failing due to missing SECRETS(MC_API), this is failing in master as well.
Please suggest the workaround to fix this.

jannisvisser:
@SwatiTh I believe we ran into this previously already as well. You need to add the same secrets as here (https://github.com/rodekruis/IBF-system/settings/secrets/actions) also in your own fork. 

Or even simpler: stop using your own fork, and simply push to branches on the upstream repository.

interfaces/IBF-dashboard/src/app/components/aggregates/aggregates.component.scss
```diff
@@ -8,6 +8,7 @@
 
   .aggregate-header-item-label {
     margin: 8px 0;
+    font-weight: 600;
```
jannisvisser, line 11: This is for the bold text, but that one is already reviewed according to DevOps, and I thought I saw it in a different PR already before?

jannisvisser (changes requested):
@SwatiTh see comments already here in this PR, even though I am counting on a new PR from upstream feature branch to upstream master.

PR #885, opened 2021-09-24
fix: increase nginx file size limit AB#9798

jannisvisser:


jannisvisser:
@gulfaraz this changes is needed for the raster-file-posting to work. Merging already because of need to test on ibf-test. But let me know if there's any issue with this.

PR #888, opened 2021-09-27
Feat.egy mock

jannisvisser:


jannisvisser:
@gulfaraz @SwatiTh I judged it was really handy to a 'mock-all' endpoint, now that we have decoupled the pipeline, because otherwise you would need to run the mock-endpoint like 8 times after every time you seed. Additionally I removed KEN (Kenya) from .env variables, as we don't have a mock for it yet, and we will simply put it back (and create mock-data) once we start working on it.

SwatiTh:
Looks Good to me

jannisvisser:
@SwatiTh thanks. I think it would be good if you're a bit more descriptive in your reviews, by describing what you tested and/or including screenshots (for the admin-boundary bug for example).

PR #893, opened 2021-10-01
Create activation report page on the dashboard AB#8538

SwatiTh:


jannisvisser (changes requested):
- change to CSV
- content should be clickable/selectable > copyable
- also not scrollable (or otherwise: not all endpoint records shown)

jannisvisser (approved):

PR #894, opened 2021-10-01
fix: chat intro text bug AB#9874

jannisvisser:


interfaces/IBF-dashboard/src/app/components/chat/chat.component.html
```diff
@@ -19,64 +19,11 @@
...
-          'chat-component.' +
-            disasterCategory +
-            'new-event-active-trigger.welcome'
-            | translate
-              : {
-                  name: authService.displayName,
-                  leadTime: eventService.state.triggerLeadTime,
-                  timeUnit: eventService.state.timeUnit,
-                  disasterTypeLabel: disasterTypeLabel,
-                  monthYear: eventService.state.firstLeadTimeName
```
jannisvisser, line 42: @SwatiTh 
- the {{startDate}} bug was startDate was missing here as a parameter that needed to be filled in
- the reason you did see it at first, but not when you checked again at a later date, is because of the distinction between 'new-event-active-trigger' and 'active-event-active-trigger', where the current date matters in relation to the event start date. And for 'active-event-active-trigger', startDate was correctly added (see below).
- Because additionally I found that both texts were now identical anyway, I decided to remove the distinction altogether. This simplifies the code.

interfaces/IBF-dashboard/src/assets/i18n/en.json
```diff
@@ -117,12 +113,8 @@
       "no-event-no-trigger": {
         "welcome": "Hello {{ name }}. To your right is the map of your country, where you can turn layers on and off. There are currently <strong>no alerts</strong> for dengue activated."
       },
-      "new-event-active-trigger": {
-        "welcome": "Hello {{ name }}. A {{ disasterTypeLabel }} alert for has been activated for <strong>{{ monthYear }}</strong>.",
-        "new-event-early-trigger": "The system however detects an alert as soon as <strong>{{ leadTime }} month(s) from now</strong>, which was not detectable earlier."
-      },
       "active-event-active-trigger": {
-        "welcome": "Hello {{ name }}. A {{ disasterTypeLabel }} alert has earlier been activated on <strong>{{ startDate }}</strong>. The trigger is activated for <strong>{{ monthYear }}</strong>."
+        "welcome": "Hello {{ name }}. A trigger warning for <strong>{{ disasterTypeLabel }}</strong> was activated on <strong>{{ startDate }}</strong>. The trigger is activated for <strong>{{ monthYear }}</strong>."
```
jannisvisser, line 117: @SwatiTh I also copied the new text structure to the separate 'dengue' and 'malaria' sections (which stupidly enough have separate sections here in en.json, while other disaster-types don't). This is unfortunately badly coded back in the day and should be improved in the future hopefully.

interfaces/IBF-dashboard/src/app/services/event.service.ts
```diff
@@ -134,18 +127,18 @@ export class EventService {
 
   private onTriggerPerLeadTime = (timesteps) => {
     let firstKey = null;
-    Object.keys(timesteps).forEach((key) => {
-      if (timesteps[key] === '1') {
-        firstKey = !firstKey ? key : firstKey;
-      }
-    });
+    Object.keys(timesteps)
+      .sort((a, b) => (a > b ? 1 : -1))
```
jannisvisser, line 131: @SwatiTh I added this 'sort' line to make sure that the timesteps are in the right order. Because I think this is again a case where the random can change randomly somehow, because in my local it (correctly) said '3 days', while on ibf-test it (incorrectly) said '7 days'. With this change I'm making sure that it is always the lowest one: '3 days'.

SwatiTh:
heavy-rain-view
malaria-View

SwatiTh:
Looks Good

PR #895, opened 2021-10-01
fix: show selected rain icon EGY AB#9883

jannisvisser:


interfaces/IBF-dashboard/src/app/components/disaster-type/disaster-type.component.ts
```diff
@@ -42,8 +42,7 @@ export class DisasterTypeComponent implements OnInit, OnDestroy {
         ({ activeTrigger }) => activeTrigger,
       );
       if (activeDisasterType) {
-        this.selectedDisasterType =
-          DisasterTypeKey[activeDisasterType.disasterType];
```
jannisvisser, line 46: @SwatiTh the problem here was that for 'heavy-rain' (unlike for all other disasters) the key of DisasterTypeKey is not the same as value (heavyRain instead of heavy-rain), so you couldn't use the code this way.

SwatiTh: Got our point @jannisvisser, Thanks

SwatiTh:
Nice, Showing selected now.
Heavy-rain-selected

PR #926, opened 2021-11-05
Feat.gherkin scenarios

jannisvisser:


features/IBF-portal-user/dashboard-page/View_user_header_section.feature
```diff
@@ -0,0 +1,60 @@
...
+    Then the user redirected to IBF-dashboard
+    And it shows "Logged in as" with login country's name 
+    And it contains login country name (by using which country user looged in as)
+    And it contains menu-icon, logo and "Logout"  and "Export View" buttons.
+    And it contains that country's logo on the right side of the dashboard
+    And it contains Logout and Export-view buttons 
+
+
+Scenario: View logout page
+    Given any logged in user
```
jannisvisser, line 17: The point of given is that you don't need to repeat if every time, if you have already put it in 'background' on top.

jannisvisser:
@SwatiTh I reversed the latest changes you made to this PR, as it was unclear to me what the purpose is. Moreover we said we would not do any more work on this within this item, but instead create a follow-up PBI. If your work still turns out to be relevant then, we can always get it again from your commit: https://github.com/rodekruis/IBF-system/pull/926/commits/7ab4d0c979e898c8ae5885f23aa3af42838caa0a

PR #932, opened 2021-11-19
Feat.new title segment design 9123+9121

SwatiTh:


jannisvisser (changes requested):
@SwatiTh first some administration:
- this PR is about the top bar / situational overview, yet you have called both branch and PR 'title segment', which is the other one.
- if there are merge conflicts, please merge master locally, that should solve it

then there are a couple of things not right yet about:
- you're missing completely the date of today (see Figma file). I would say to keep the format of the today-button as it is now though (and don't split it in 2 buttons)
- The refinement mentions that the width of this bar should be narrowed to only cover the chat-section (and with that, the timeline-component should move to the left)
- Why is the TODAY thing clickable and getting a color upon hovering? It should not.
- I see that you made the background white now, which indeed aligns with the Figma-file. I however remember from the meeting that we discussed not to do that actually (but keep it the way it was). However, it's fine to do it anyway, but then we should have some delineation with the title section (as is also in the Figma-file), so that it is not one big white space.
image

jannisvisser:
@SwatiTh see my changes here. The most important one was that you had pulled the timeline-section within the top bar. This was not correct. The timeline-section only moved to the left but is starting at the same place as the middle column.

I have left one task open for you, which involves the 'shadow' thing. Namely, with the shadow, the height of the bar gets bigger too, and thereby bigger than the timeline-section (but even apart from that, the height should not get more at all). Can you find a fix for that?

jannisvisser:
closed as integrated in #931

PR #935, opened 2021-11-22
Feat.show not all areas typhoon

jannisvisser:


jannisvisser:
@SwatiTh again already merging this, so phuoc can access it. You can look at it though (you should see that in trigger-mode, not the whole country loads any more, but just a portion).

PR #945, opened 2021-12-03
feat: show multiple events AB#10943

jannisvisser:


services/API-service/src/scripts/scripts.service.ts
```diff
@@ -197,24 +216,31 @@ export class ScriptsService {
...
       return null;
     }
   }
 
-  private getTyphoonLeadTime(): string {
-    return LeadTime.hour72;
+  private getTyphoonLeadTime(eventNr: number = 1): string {
+    if (eventNr === 1) {
+      return LeadTime.hour72;
+    } else if (eventNr === 2) {
```
SwatiTh, line 238: Why are we hard coding eventNr values here? what if we have more number of events

services/API-service/src/api/metadata/metadata.service.ts
```diff
@@ -1,5 +1,6 @@
 import { Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
+import { eventNames } from 'process';
```
SwatiTh, line 3: I think we should remove this if we are not using it somewhere else

jannisvisser:
@SwatiTh so also merging this one (your commit looks good)

PR #948, opened 2021-12-10
feat: TRIGGERED NON TRIGGER TITLE STATE

SwatiTh:


jannisvisser (approved):
@SwatiTh I only cleaned up some unused icons & updated icon-names

PR #958, opened 2021-12-20
feat: update login screen

SwatiTh:


jannisvisser (changes requested):
@SwatiTh 
- this item also has to include updating the IBF-logo to the colored one (see screenshot in main item). 
- you now use the GET countries-endpoint to determine which icons to show at all. This is incorrect however. You should always show all known disaster-types (so base that on DISASTER_TYPES_SVG_MAP I guess?) and use the countries-endpiont to determine which icons to show as navy (the rest is then grey). In fact I'm not sure what role 'this.country' should play at all in login.page.ts? There is no country selected right? So it should not be there at all.
- I would expect the icons to be a bit more centered vertically somehow, it feels a bit off now. What do you think?
- The cursor should not switch to a button/hand icon when hovering over it, because it's not a button, which you might think this way.
- It would however be good maybe to have a 'title' tooltip appearing when you hover over with the name of the disaster-type. Could be considered nice-to-have though, not must-have.
- Dengue and Malaria have the same icon, should we deduplicate it? (I guess fine for now though.)

SwatiTh:
this item also has to include updating the IBF-logo to the colored one (see screenshot in main item).

- Done

you now use the GET countries-endpoint to determine which icons to show at all. This is incorrect however. You should always show all known disaster-types (so base that on DISASTER_TYPES_SVG_MAP I guess?) and use the countries-endpiont to determine which icons to show as navy (the rest is then grey). In fact I'm not sure what role 'this.country' should play at all in login.page.ts? There is no country selected right? So it should not be there at all.

- As per current implementation, all the icons will be in navy-blue colour in case of staging and local environment where the country is not selected or null. In production env. it will behave as expected.

I would expect the icons to be a bit more centered vertically somehow, it feels a bit off now. What do you think?
The cursor should not switch to a button/hand icon when hovering over it, because it's not a button, which you might think this way.

- Done

It would however be good maybe to have a 'title' tooltip appearing when you hover over with the name of the disaster-type. Could be considered nice-to-have though, not must-have.

- We don't have tool-tip as of now in the application. Not implementing as of now.

Dengue and Malaria have the same icon, should we deduplicate it? (I guess fine for now though.)

- Fine for now

PR #959, opened 2021-12-20
feat: added notification if login fails AB#11318

SwatiTh:


interfaces/IBF-dashboard/src/app/auth/auth.service.ts
```diff
@@ -112,7 +114,12 @@ export class AuthService implements OnDestroy {
     this.router.navigate(['/']);
   };
 
-  private onLoginError = (error) => {
+  private onLoginError = async ({ error }) => {
+    const toast = await this.toastController.create({
+      message: `Authentication Failed: ${error?.message}`,
+      duration: 2000,
```
jannisvisser, line 120: I think this can be a bit longer, especially because the toast is not very visually popping out, so the user might need a second to locate it. I would put it at 5 seconds maybe?

interfaces/IBF-dashboard/src/app/auth/auth.service.ts
```diff
@@ -112,7 +114,12 @@ export class AuthService implements OnDestroy {
     this.router.navigate(['/']);
   };
 
-  private onLoginError = (error) => {
+  private onLoginError = async ({ error }) => {
+    const toast = await this.toastController.create({
+      message: `Authentication Failed: ${error?.message}`,
```
jannisvisser, line 119: Is 'Unauthorized' clear enough? Or should we translate this to something like 'email and/or password unknown'? (so do that if error.status === 401, for other cases show error.message like now)

jannisvisser (changes requested):
@SwatiTh looks good, just some small comments.

jannisvisser (approved):
@SwatiTh looks good from code :) I have no time to test it, but I trust it can be merged.

PR #971, opened 2022-01-10
Feat.upload static data

jannisvisser:


jannisvisser:
@arsforza FYI: this involves another addition to the test scenarios. Just good to be aware.

PR #974, opened 2022-01-10
fix: pass support email as var to dashboard AB#11841

jannisvisser:


jannisvisser:
@arsforza I strongly think that this was the issue.

PR #981, opened 2022-01-17
feat: return column headers ico empty activation log AB#11830

jannisvisser:


jannisvisser:
merged this into #983

PR #987, opened 2022-01-24
Fix.login icons

jannisvisser:


jannisvisser:
@arsforza already merging, so that it's ready for sprint review. But please review nonetheless (you can do that on ibf-test though then).

PR #1019, opened 2022-02-18
fix: Allow uppercase characters in email login field AB#12372

arsforza:


services/API-service/src/api/user/user.service.ts
```diff
@@ -41,7 +41,8 @@ export class UserService {
 
   public async create(dto: CreateUserDto): Promise<UserResponseObject> {
     // check uniqueness of username
-    const { email, password } = dto;
+    let { email, password } = dto;
```
jannisvisser, line 44:
```suggestion
    const { email, password } = dto;
```

jannisvisser (changes requested):

PR #1120, opened 2022-06-24
update popup PHP/ETH

aklilu:


aklilu:
also updated Ethiopia drought popup info

PR #1200, opened 2022-10-04
update pop up text - add text for affected_population layer

aklilu:
@jannisvisser @arsforza updated pop up text for affected_population layer Philippiness


arsforza:
@aklilu I copied your changes in another PR because this had conflicts with master.

I saw that you added a new row for the layer, in position 475, but the layer was already in the file, on row 40 (I copied your changes there).
The new layers are always already in the xlsx, since they're added by @jannisvisser and me when we prepare the file.

PR #1203, opened 2022-10-07
fix: show admin-popup after zoom AB#15614

jannisvisser:


interfaces/IBF-dashboard/src/app/components/disaster-type/disaster-type.component.ts
```diff
@@ -56,7 +56,7 @@ export class DisasterTypeComponent implements OnInit, OnDestroy {
       this.disasterTypesCounter = 0;
       this.disasterTypes = country.disasterTypes;
       this.disasterTypes.sort((a, b) =>
-        a.disasterType > b.disasterType ? 1 : -1,
+        a.disasterType < b.disasterType ? 1 : -1,
```
jannisvisser, line 59: forget to turn it back? ;)

PR #1218, opened 2022-10-21
fix: process alert_threshold correctly AB#15876

jannisvisser:


services/API-service/src/api/event/event.service.ts
```diff
@@ -606,14 +609,16 @@ export class EventService {
...
       );
-      eventArea.activeTrigger = true;
-      eventArea.endDate = endDate;
-      if (affectedArea.triggerValue > 0) {
-        eventArea.thresholdReached = true;
-        idsToUpdateAboveThreshold.push(eventArea.eventPlaceCodeId);
-      } else {
-        eventArea.thresholdReached = false;
-        idsToUpdateBelowThreshold.push(eventArea.eventPlaceCodeId);
+      if (affectedArea) {
```
jannisvisser, line 612: @arsforza this line was the fix (the rest is just other small refactoring I noticed). The way to reproduce it locally was to first have triggered areas already in the database (using the mock endpoint) and then uploading phuoc's json.

PR #1485, opened 2024-03-11
AB#26410 Make port ibf service port configurable locally and update setup Readme

RubenGeo:


example.env
```diff
@@ -11,6 +11,8 @@ DB_PASSWORD=
 SECRET=
 RESET_SECRET=
 NODE_ENV=
+LOCAL_PORT_IBF_SERVICE=4000 # If you do not specify a port, the default port is 3000. this only applies to the override docker-compose file
```
RubenGeo, line 14: If we merge this PR you need to set this in your local .env @arsforza @jannisvisser

PR #1487, opened 2024-03-18
AB#26925 Sync stores and layers to geoserver

RubenGeo:


jannisvisser:
@RubenGeo closing because this was already integrated in the other branch.

PR #1494, opened 2024-05-06
fix(security): remove crypto and unused packages AB#27340

gulfaraz:


RubenGeo:
@gulfaraz I see an import of crypto here https://github.com/rodekruis/IBF-system/blob/build.crypto/services/API-service/src/api/user/user.entity.ts#L11 and it is being used here https://github.com/rodekruis/IBF-system/blob/build.crypto/services/API-service/src/api/user/user.entity.ts#L91. Will this change not break this? Or is this a different crypto?

gulfaraz:
> @gulfaraz I see an import of crypto here https://github.com/rodekruis/IBF-system/blob/build.crypto/services/API-service/src/api/user/user.entity.ts#L11 and it is being used here https://github.com/rodekruis/IBF-system/blob/build.crypto/services/API-service/src/api/user/user.entity.ts#L91. Will this change not break this? Or is this a different crypto?

I assumed it would break that line too. But it did not. Turns out that there is an internal node crypto package. I tested it but it would be nice to have it confirmed.

RubenGeo (approved):
I checked out to this branch ran `npm i`, created a new user and logged in succesfully. Also node's documentation confirms it's a default package, so all good :)

PR #1498, opened 2024-05-14
fix empty timestamp and reduce html duplication

gulfaraz:


arsforza (commented):
@gulfaraz we are not using timestamp anywhere
I can approve the changes but we need a new item to pass the timestamp to dialogue-turn

PR #1590, opened 2024-09-19
Add Playwright with eslint and example login test

Piotrk39:
## Describe your changes

Adds Playwright E2E automation tool to IBF-system Github repository, that includes:
- E2E test Directory
- E2E example login test
- Page Object Model (POM)
- Eslint for E2E test directory

ToDo:
- [x] Connect with API tests to generate token and reset DB for creating correct state
- [x] Remove Cypress and it's dependencies
- [x] Add Readme
- [x] Add CI/CD configuration to run on every PR


e2e/tests/Login/SuccessfullyLogIn.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+import {
+  getAccessToken,
+  resetDB,
+} from '../../../services/API-service/test/helpers/utility.helper';
+import LoginPage from '../../Pages/LoginPage';
+
+let accessToken: string;
+
+test.beforeEach(async ({ page }) => {
+  accessToken = await getAccessToken();
```
jannisvisser, line 12: It seems duplicate that you first log in already via API here, and then again via portal. In fact, resetting DB is not needed for this particular front-end login test.

e2e/tests/Login/SuccessfullyLogIn.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+  await resetDB(accessToken);
+
+  // Login
+  const loginPage = new LoginPage(page);
+  await page.goto('/');
```
jannisvisser, line 17: This is failing for me. When I replace it by 'http://localhost:4200' it works.

e2e/tests/Login/SuccessfullyLogIn.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+  await loginPage.login(
+    process.env.USERCONFIG_MALAWI_EMAIL,
```
jannisvisser, line 19: 1. I don't think we should load this via env. Instead can we load it from 'users.json' somehow? Or in the same way the utility.helper loads it for API-service? Let's discuss.
2. There's no particular reason to use malawi-specific email. We have a multi-country admin-user 'dunant@redcross.nl' which should be used by default I think. Only to deviate from in specific test cases.

e2e/tests/Login/SuccessfullyLogIn.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+    process.env.USERCONFIG_MALAWI_PASSWORD,
+  );
+});
+
+test('Successfully Login', async ({ page }) => {
+  await page.waitForURL((url) => url.pathname.startsWith('/'));
```
jannisvisser, line 25: Is this assertion actually correct? When I get the test running and then  print out the 'url' variable here, it shows me that ur.pathname = '/login', which passes the assertion but is not actually what we want to test right? We want to test that it goes from the login page to the dashboard page.

e2e/playwright.config.ts
```diff
@@ -0,0 +1,78 @@
...
+  /* Run tests in files in parallel */
+  fullyParallel: false,
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
+  /* Retry on CI only */
+  retries: 1,
+  /* Opt out of parallel tests on CI. */
+  reporter: [
+    ['list'],
+    // [
```
jannisvisser, line 21: can we remove this (and check for more?) commented code?

.github/workflows/test_e2e_ibf_system.yml
jannisvisser, line 1: The automatic test in the PR is failing for now still. Do you need help fixing that?

jannisvisser (changes requested):
@Piotrk39 good start!
- see specific comments in PR. Let's discuss some of those in our call already.
- additionally, can you add a basic readme?

e2e/tests/Login/SuccessfullyLogIn.spec.ts
Piotrk39: 1. We can surely use the JSON to load those users if it is more convenient.
2. The reason Malawi was used is that it was so far the only setup that shows correct data in my local environment. For the login test of course it does not matter.

Note:

Tests that will run in the future should probably run per project so different login users should be used. I did not dive so much into that yet but it is food for thought.

e2e/tests/Login/SuccessfullyLogIn.spec.ts
Piotrk39: As discussed the test will assert that user was redirected to base URL.

e2e/playwright.config.ts
Piotrk39: So as I mentioned the commented out code is the code used for reporting the tests status to Azure or any other Test management tool. For this reason I would like to keep it and use it uncommented when needed. For most part it will stay commented since we don't want a report to be sent per each commit to PR.

Piotrk39:
> @Piotrk39 good start!
> 
> * see specific comments in PR. Let's discuss some of those in our call already.
> * additionally, can you add a basic readme?

@jannisvisser  I will add the readme as the last step indeed. Forgot to put it in my ToDo list. Thanks

e2e/tests/Login/SuccessfullyLogIn.spec.ts
jannisvisser: Remove need for BASE_URL and BASE_ (the other one).

e2e/tests/Login/SuccessfullyLogIn.spec.ts
jannisvisser: Use dunant user, and get it from users.json if that's possible, otherwise recreate in e2e folder?

Piotrk39: Done

Piotrk39:
@jannisvisser I have added a simple README to the branch and it is located in ./e2e directory.

jannisvisser (approved):

.github/workflows/test_e2e_ibf_system.yml
```diff
@@ -0,0 +1,97 @@
...
+
+      - name: Lint E2E-Tests code
+        working-directory: ${{ env.e2eTestsPath }}
+        run: 'npm run lint'
+
+      - name: Set ENV-variables for test-environment
+        run: |
+          cp example.env .env
+
+      - name: Install API-service code-dependencies
```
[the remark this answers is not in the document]
jannisvisser: @gulfaraz good question. It seemed necessary, because the e2e tests are run from the host with 'npm test', and they in turn use some of the API-service files (e.g. utility.helper.ts) to reset the DB, which in turn needs the 'supertest' package. And without this 'npm install' step on the host instead of within the container, this package was not found.

Do you think this setup should be changed?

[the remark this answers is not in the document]
gulfaraz: It shouldn't be that the installation is needed on the host machine while the service is running inside docker.

I will investigate this before I propose a change.

gulfaraz:
@Piotrk39 is there a benefit to keep the e2e tests in a separate workflow? I would like a setup where a failure of an end-2-end test will block the [version bump](https://github.com/rodekruis/IBF-system/blob/502618ba423dd6498b11fedb9379b68500590c7f/.github/workflows/workflow.yml#L82-L88) in the other workflow. As these workflows are currently independent, the version gets bumped regardless of the e2e tests.

FYI @jannisvisser

Piotrk39:
@gulfaraz We should add new item to fix that 

> @Piotrk39 is there a benefit to keep the e2e tests in a separate workflow? I would like a setup where a failure of an end-2-end test will block the [version bump](https://github.com/rodekruis/IBF-system/blob/502618ba423dd6498b11fedb9379b68500590c7f/.github/workflows/workflow.yml#L82-L88) in the other workflow. As these workflows are currently independent, the version gets bumped regardless of the e2e tests.
> 
> FYI @jannisvisser

jannisvisser:
@gulfaraz could we alternatively block merging to master if not all tests succeed?

gulfaraz:
Yes, if we can figure out a way to link the other workflow to the [version bump](https://github.com/rodekruis/IBF-system/blob/502618ba423dd6498b11fedb9379b68500590c7f/.github/workflows/workflow.yml#L82-L88) step.

It also doesn't make sense to me that we repeat certain steps in both workflows. Github actions has a limit so I think it's wise to use this limited resource efficiently. But if the need is there we can absolutely repeat certain steps, this is why I asked if the decision to have a separate workflow is deliberate.

Issue #1515, opened 2024-07-18
[Task] Switch to mjml to handle email templates


arsforza:
### **Who requested this feature?**
@gulfaraz 

### **User Story**
As a developer
I want to use mjml to create email templates
So that emails are easier to create and style

### **What needs to be done**
- Install mjml
- Migrate to object structure
- Test for multiple disasters

### After first review with @gulfaraz 
- [x] Fix width on mobile
- [x] Set Arial as font
- [x] Add margin between action button and description on mobile
- [x] Fix "from nows from now" typo
- [x] Use correct button labels
- [x] Add Mailchimp footer
- [x] Fix event trigger issued text's size and spacing
- [ ] Use `date-fns` instead of `moment`
- [ ] (optional) Fix rounded buttons on Outlook (Windows)
- [ ] Make single-logo countries logos smaller (by limiting height and adjusting width to that?)

jannisvisser:
@arsforza @gulfaraz 
- I have merged the fix.email-logos branch into this one as discussed yesterday.
- I noticed that the single logo countries have a huge logo, but that was already the case before, so decided to not let it block the email logo item, but maybe good to fix in this item still? (made a task above for  it)
![Image](https://github.com/user-attachments/assets/7fc93941-58fa-4a0d-bf0c-32f03d7f560b)

gulfaraz:
> I noticed that the single logo countries have a huge logo, but that was already the case before, so decided to not let it block the email logo item, but maybe good to fix in this item still? (made a task above for it)

You should use `height: 50px` in the [CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/height) or image `height="50"` in the [IMG](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/height) tag to set the height of the logo.

gulfaraz:
The can be closed after merging PRs #1527 and #1594.

PR #1656, opened 2024-10-15
fix: compact number in email should consider format

gulfaraz:
## Describe your changes

This PR is a result of [feedback](https://github.com/rodekruis/Anticipatory-Action/issues/880) from @gal-agmon on #1515

Compact numbers in email consider number format. This changes PHL typhoon emails,

Before this PR, https://us18.campaign-archive.com/?u=e71f3b134823403aa6fe0c6dc&id=ce2d4c8cfb
After this PR, https://us18.campaign-archive.com/?u=e71f3b134823403aa6fe0c6dc&id=44951c89fb

### DIFF
--- a/services/API-service/src/api/notification/notification-content/notification-content.service.ts
+++ b/services/API-service/src/api/notification/notification-content/notification-content.service.ts
@@ -4,6 +4,7 @@ import { InjectRepository } from '@nestjs/typeorm';
 import { Repository } from 'typeorm';
 
 import { EventSummaryCountry, TriggeredArea } from '../../../shared/data.model';
+import { NumberFormat } from '../../../shared/enums/number-format.enum';
 import { HelperService } from '../../../shared/helper.service';
 import { LeadTime } from '../../admin-area-dynamic-data/enum/lead-time.enum';
 import { CountryEntity } from '../../country/country.entity';
@@ -194,7 +195,11 @@ export class NotificationContentService {
       disasterType,
     );
 
-    data.totalAffectedOfIndicator = this.getTotalAffected(data.triggeredAreas);
+    const indicatorMetadata = await this.getIndicatorMetadata(disasterType);
+    data.totalAffectedOfIndicator = this.getTotal(
+      data.triggeredAreas,
+      indicatorMetadata.numberFormatMap,
+    );
     data.eapAlertClass = event.disasterSpecificProperties?.eapAlertClass;
     return data;
   }
@@ -261,12 +266,22 @@ export class NotificationContentService {
     });
   }
 
-  private getTotalAffected(triggeredAreas: TriggeredArea[]) {
-    return parseFloat(
-      triggeredAreas
-        .reduce((acc, triggeredArea) => acc + triggeredArea.actionsValue, 0)
-        .toFixed(2),
+  private getTotal(
+    triggeredAreas: TriggeredArea[],
+    numberFormat: NumberFormat,
+  ) {
+    const total = triggeredAreas.reduce(
+      (acc, { actionsValue }) => acc + actionsValue,
+      0,
     );
+
+    if (numberFormat === NumberFormat.perc) {
+      // return average for percentage
+      return total / triggeredAreas.length;
+    }
+
+    // return sum
+    return total;
   }
 
   private async getFirstLeadTimeDate(
--- a/interfaces/IBF-dashboard/src/app/pipes/compact.pipe.ts
+++ b/interfaces/IBF-dashboard/src/app/pipes/compact.pipe.ts
@@ -28,7 +28,8 @@ export class CompactPipe implements PipeTransform {
     }
 
     const style = format === NumberFormat.perc ? 'percent' : 'decimal';
-    const maximumSignificantDigits = value > 100 ? 2 : 1;
+    const maximumSignificantDigits =
+      value > 100 || format === NumberFormat.perc ? 2 : 1;
 
     let min = 0;
     let prefix = '';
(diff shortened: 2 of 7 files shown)


jannisvisser:
@gulfaraz so from 10% onwards it's rounded to tens of percentages? This is not what you said in the meeting, nor what Gal Said in the meeting right?

jannisvisser:
@gulfaraz also, this is not correct in the new email
image

gulfaraz:
@jannisvisser Thanks! Not much was said in the meeting. These changes are based on a review from weeks ago. I'll try to find some notes and update this PR.

services/API-service/src/api/notification/notification-content/notification-content.service.ts
```diff
@@ -261,12 +266,22 @@ export class NotificationContentService {
...
+    numberFormat: NumberFormat,
+  ) {
+    const total = triggeredAreas.reduce(
+      (acc, { actionsValue }) => acc + actionsValue,
+      0,
     );
+
+    if (numberFormat === NumberFormat.perc) {
+      // return average for percentage
+      return total / triggeredAreas.length;
```
jannisvisser, line 280: @gulfaraz this is not fully correct. You have to take the weighted average, in this case through 'weightVar=total_houses', which is also an indicator-attribute. In aggregates.service in the front-end there is an example of how this is used. You can maybe argue this is already to the sum we had before, but this will lead to differing numbers between email and portal.

jannisvisser: Please realize that having this total figure in the email is a new feature, that is not currently not on production. So we're losing nothing with just omitting it. 

Pseudocode formula is just something like: total-% = (area1-%*area1-total_houses + area2-%*area2-total_houses) / (area1-total_houses = area2-total_houses).

You probably don't have this additional data on 'total_houses' per area readily available at the place where you need it.

gulfaraz:
The indicator value shown in the email event is incorrect. See review [here](https://github.com/rodekruis/IBF-system/pull/1656#discussion_r1803015897) and [IBF Dev Chat](https://teams.microsoft.com/l/message/19:bdb9d621f18c48c7917e545bf6d7aabe@thread.v2/1728991862942?context=%7B%22contextType%22%3A%22chat%22%7D) for details.

> In our emails we used to show the total indicator as % of houses exposed: Approximately 7,23 % of houses exposed
> This is a total of all the percentage values [0.53, 0.51,  0.5, 0.47, 0.45, 0.39,  0.3, 0.3,  0.3, 0.26, 0.25, 0.17, 0.17, 0.16, 0.16, 0.14, 0.14, 0.12, 0.12, 0.12, 0.12, 0.11, 0.11, 0.11, 0.11, 0.11, 0.11, 0.08, 0.07, 0.07, 0.06, 0.05, 0.05, 0.05, 0.05, 0.03, 0.03, 0.03, 0.03, 0.03, 0.03, 0.03, 0.02, 0.02, 0.01, 0.01, 0.01, 0.01, 0.01, 0.01, 0.01, 0.01, 0.01, 0.01, 0.01, 0.01, 0.01, 0.01, 0.01, 0.01] = 7.229999999999999
> 
> Which should be interpreted as 723%.

The old indicator is wrong and there is no easy way to calculate the correct value without significant refactoring. So I removed the indicator following @jannisvisser's advice. After this change, the emails look like,
1. PHL typhoon https://us18.campaign-archive.com/?u=e71f3b134823403aa6fe0c6dc&id=d31bea6478
2. PHL flood https://us18.campaign-archive.com/?u=e71f3b134823403aa6fe0c6dc&id=591b54ea33
3. UGA flood https://us18.campaign-archive.com/?u=e71f3b134823403aa6fe0c6dc&id=625a1ea130

services/API-service/src/api/notification/notification-content/notification-content.service.ts
gulfaraz: > To be honest I was actually meaning to only omit the line in the percentage case.

This wasn't clear before and is now perplexing. Did the feature requirements describe this exception?

> Because it does work correctly in the sum-case, which is the vast majority of all implementations (population_affected).

We [agreed to have shared behaviour across countries and hazards](https://teams.microsoft.com/l/message/19:ad7e3a0c71ad4e9ba5b84a799eb507db@thread.v2/1728055901400?context=%7B%22contextType%22%3A%22chat%22%7D) without exceptions like these. I understand we implemented this feature before our agreement, but we're deciding to sign off on the feature now.

The requirement is to show the same aggregate value in the email as in the dashboard. If the suggestion is to show a value in the dashboard but not in the email, _I advise to not ship a partial implementation_. Let's refine this feature when it returns as feedback.

> But in this particular case, it might be better to split it with a clear comment in the code explaining why.

This addresses the discrepancy for us devs. However, it does not explain why the value is missing to the end-user, those demoing IBF, design, and other stakeholders.

We have the opportunity to,
1. Avoid the effort to document this exception
2. Avoid communicating this exception to the relevant people mentioned above
3. Avoid a future refactor to remove this exception

jannisvisser: Ok, I'm convinced by your argumentation, but do not appreciate the unfriendly tone of the choice of the word 'perplexing' (unless I'm misinterpreting somehow).

gulfaraz: By perplexing, I meant **confusing**. I apologize if it came across as unfriendly. It was very neutral.

jannisvisser: Ok, clear, thanks :)

jannisvisser (approved):

PR #1662, opened 2024-10-21
Add e2e test for map functionalities in no-trigger mode

Piotrk39:
## Describe your changes

Adds e2e test for map functionalities in no-trigger mode

### DIFF
--- a/tests/e2e/tests/AggregatesSection/AggregatesHasDynamicTitle.spec.ts
+++ b/tests/e2e/tests/AggregatesSection/AggregatesHasDynamicTitle.spec.ts
@@ -45,7 +45,7 @@ test(
     await dashboard.navigateToFloodDisasterType();
     // Assertions
     await aggregates.aggregateComponentIsVisible();
-    await map.turnOffLayer({ layerName: 'Glofas stations' });
+    await map.clickLayerCheckbox({ layerName: 'Glofas stations' });
     await map.assertAggregateTitleOnHoverOverMap();
   },
 );
--- a/tests/e2e/tests/AggregatesSection/AggregatesHasDynamicTitle.spec.ts
+++ b/tests/e2e/tests/AggregatesSection/AggregatesHasDynamicTitle.spec.ts
@@ -33,7 +33,10 @@ test.beforeEach(async ({ page }) => {
     NoTriggerDataSet.UserPassword,
   );
 });
-
+// The test is skipped because of the bug that was identified during writing of this test
+// The bug is that the marker of glofas stations cannot be disabled with the first chebox click (needs several) and it is failing on flood disaster type
+// https://github.com/rodekruis/IBF-system/issues/1657
+// When the bug is fixed, the test should be unskipped
 test.skip(
   qase(
     12,
--- a/tests/e2e/tests/AggregatesSection/AggregatesHasDynamicTitle.spec.ts
+++ b/tests/e2e/tests/AggregatesSection/AggregatesHasDynamicTitle.spec.ts
@@ -35,7 +35,10 @@ test.beforeEach(async ({ page }) => {
 });
 
 test.skip(
-  qase(12, 'Aggregates title should be dynamic no-trigger'),
+  qase(
+    12,
+    'Aggregates title should be dynamic upon hovering over map district',
+  ),
   async ({ page }) => {
     const dashboard = new DashboardPage(page);
     const aggregates = new AggregatesComponent(page);
--- a/tests/e2e/tests/AggregatesSection/AggregatesHasDynamicTitle.spec.ts
+++ b/tests/e2e/tests/AggregatesSection/AggregatesHasDynamicTitle.spec.ts
@@ -34,7 +34,7 @@ test.beforeEach(async ({ page }) => {
   );
 });
 
-test(
+test.skip(
   qase(12, 'Aggregates title should be dynamic no-trigger'),
   async ({ page }) => {
     const dashboard = new DashboardPage(page);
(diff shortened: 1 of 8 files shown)


tests/e2e/tests/AggregatesSection/AggregatesHasDynamicTitle.spec.ts
```diff
@@ -34,7 +34,7 @@ test.beforeEach(async ({ page }) => {
   );
 });
 
-test(
+test.skip(
   qase(12, 'Aggregates title should be dynamic no-trigger'),
```
jannisvisser, line 38: @Piotrk39 I think this description can be a bit more detailed. The point of the test is that when you hover over the map, you see the title changing right? Can you include that a bit more?

tests/e2e/tests/AggregatesSection/AggregatesHasDynamicTitle.spec.ts
```diff
@@ -45,7 +45,7 @@ test(
     await dashboard.navigateToFloodDisasterType();
     // Assertions
     await aggregates.aggregateComponentIsVisible();
-    await map.turnOffLayer({ layerName: 'Glofas stations' });
+    await map.clickLayerCheckbox({ layerName: 'Glofas stations' });
```
jannisvisser, line 48: so the point why this isn't working is that the glofas stations layer does not toggle off when you click it? And subsequently when hovering over the map, it doesn't do that correctly? Why actually, as a user myself I could still hover over admin areas when the glofas station layer is on right? (let's indeed call to discuss further if needed)

tests/e2e/tests/AggregatesSection/AggregatesHasDynamicTitle.spec.ts
```diff
@@ -45,7 +45,7 @@ test(
...
     await map.assertAggregateTitleOnHoverOverMap();
```
jannisvisser, line 49: Another question: is it actually the best choice to put this in the file/folder on 'verifying aggregates section', as the functionality we test is actually in the map (hovering), the consequences just happen to be in other sections. Just wondering if you have an opinion on this. (Depending on the answer, there might be more occurences of this.)

tests/e2e/tests/Map/AssertMapFunctionalitiesNotrigger.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  await page.goto('/');
+  await loginPage.login(
+    NoTriggerDataSet.UserMail,
+    NoTriggerDataSet.UserPassword,
+  );
+});
+
+test(
+  qase(7, 'Verify Map functionality for no-triggered mode'),
```
jannisvisser, line 39: One thing I'm wondering about in general is how it is made clear what is actually tested in a test (and thereby what is not). In this case it is relatively clear by just looking at the code, but in some other cases it is sometimes less clear. For example in the aggregates-section above, there is much more you could potentially look at right? The numbers also changing on hover. Whether the numbers change to the right amount. Etc. Should we write this down in more detail, including maybe the things that are not tested yet? Do you have an opinion?

jannisvisser (changes requested):

tests/e2e/tests/AggregatesSection/AggregatesHasDynamicTitle.spec.ts
Piotrk39: I changed it I hope it is clearer now. I also removed no-trigger part in the end since I believe it is true no matter the state.

tests/e2e/tests/AggregatesSection/AggregatesHasDynamicTitle.spec.ts
Piotrk39: This is type of workaround that you have to perform while working with tools like Playwright, Cypress etc. Still when doing it manually I can see the title change but unfortunately when hovering with automation tool in the middle of the div or layer there are objects that interfere with this behaviour and that is why you'd like to turn off the layer (that is not tested in this case) to test the dynamic changes of aggregates header/title.

tests/e2e/tests/AggregatesSection/AggregatesHasDynamicTitle.spec.ts
Piotrk39: This can be a part of wider discussion. I believe tho in this case we are testing the Aggregates dynamic title/header and the map is strictly a trigger for asserting the correct behaviour. As reference in 121 vast majority of the test happens on the "Table" but the outcome and the test object lies somewhere else eg. "PA overview".

tests/e2e/tests/Map/AssertMapFunctionalitiesNotrigger.spec.ts
Piotrk39: I could improve on that. Finding clear titles on rather bigger tests is at times tricky.

tests/e2e/tests/AggregatesSection/AggregatesHasDynamicTitle.spec.ts
jannisvisser: ok GEFN!

tests/e2e/tests/AggregatesSection/AggregatesHasDynamicTitle.spec.ts
jannisvisser: Can you maybe add a note in the test explaining why it's skipped and what's needed to resolve it (maybe even link to specific GH issue?)

tests/e2e/tests/AggregatesSection/AggregatesHasDynamicTitle.spec.ts
jannisvisser: yes!

tests/e2e/tests/AggregatesSection/AggregatesHasDynamicTitle.spec.ts
Piotrk39: Done

Piotrk39:
@jannisvisser you can do a final review

jannisvisser (approved):

PR #1679, opened 2024-10-28
Fix.filter area list exposed only

jannisvisser:
## Describe your changes

- Resolves [issue 889](https://github.com/rodekruis/Anticipatory-Action/issues/889). In the end by filtering in the front-end, similar to how the email list is filtered. 
- This solves the typhoon issue of showing 1647 areas for warning events.
- For floods, we're accepting for now that this means that the map will still show more areas potentially than the area list. At least the area list coincides with the email.
- Additionally, calculate aggregate count based on the filtered array, not on separate backend property.
- Unrelated, but remove outdated split by alertClass in area-list. 

## Notes for the reviewer

### DIFF
--- a/interfaces/IBF-dashboard/src/app/components/chat/chat.component.html
+++ b/interfaces/IBF-dashboard/src/app/components/chat/chat.component.html
@@ -65,21 +65,19 @@
       @if (
         filteredStoppedAreas.length === 0 && filteredActiveAreas.length === 0
       ) {
-        @if (activeAreas.length) {
-          <app-event-speech-bubble
-            [type]="'active'"
-            [event]="event"
-            [selectedEvent]="eventState?.event?.eventName"
-            [disasterTypeLabel]="disasterType?.label"
-            [disasterTypeName]="disasterType?.disasterType"
-            [clearOutMessage]="getClearOutMessage(event)"
-            [countryCodeISO3]="country?.countryCodeISO3"
-            [areas]="activeAreas"
-            [adminAreaLabelPlural]="adminAreaLabelPlural"
-            [actionIndicatorLabel]="actionIndicatorLabel"
-            [actionIndicatorNumberFormat]="actionIndicatorNumberFormat"
-          ></app-event-speech-bubble>
-        }
+        <app-event-speech-bubble
+          [type]="'active'"
+          [event]="event"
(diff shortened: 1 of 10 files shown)


jannisvisser:
@gulfaraz see latest commit, which in the end filters in the front-end on actionsValue > 0 (to filter out PHL areas) OR on triggerValue > 0 (to make sure UGA flood warning areas are kept in, which I realized later was necessary after all).

This also means that in the end the PHL flood production observation of 2 areas showing  stays complete as is for now (both in map and in list). But that is fine, we can simply look at it in the [separate bug](https://github.com/rodekruis/Anticipatory-Action/issues/891) there already is for that.

I also realize this changes the logic slightly compared to the email, which does not have this OR-filter. But in practice, this was already working a bit different anyway, where the count of warning areas was shown, but the area-list itself was not, unlike the chat.
image

jannisvisser:
@gulfaraz (or @arsforza depending on who ends up reviewing this). Since our last conversation I've ended up needing to solve things quite  differently again. Here an attempt to structure.
- 1st new commit: small changes to make sure scenario with 0 exposed works (no specific complexities here)
- 2nd new commit: then, I realized something was going wrong with the earlier changes on the aggregate count per event. Namely, in national view multi-event it was fine, but after selecting 1 event, the count for the other 2 events went to 0.
  - this is because the count was based in front-end on triggeredAreas.length, whereas this array after event-selection is filtered on event-aras only, that is why Andrea put in the event-summary call I understand now.
  - but when reverting that, I realized that I had to apply the same filter as in the front-end (actionsValue > 0 OR triggerValue > 0) then also in this back-end call, which felt really wrong (having this same logic in 2 places, of which one in front-end, and one in back-end)
- so (3rd new commit) I went back to the drawing board a little, and I realized I could solve this after all in the GET triggered-areas endpoint, by putting this filter in the event-query there. Which comes then in the place of this complicated logic of having an optional IN-clause, (which is partly legacy from a time when triggerValue was not yet an attribute of event-place-code I think)
 - still the same filter is needed on GET events, but at least this is now at 2 places in the same service in the back-end, which is a lot more contained.
- 4th new commit: this subsequently allows to remove the filter in the front-end again
  - and to remove any filtering in the email on this 
- I've tested emails/portal for all disaster-types

arsforza (approved):

PR #1730, opened 2024-10-29
test: add map service unit tests

arsforza:
## Describe your changes

## Notes for the reviewer


interfaces/IBF-dashboard/src/app/services/map.service.ts
```diff
@@ -1008,7 +1003,7 @@ export class MapService {
     if (
       !area.triggerValue ||
       area.triggerValue < 1 ||
-      !this.eventState?.event?.thresholdReached
```
jannisvisser, line 1011: @arsforza are you sure this is OK. I feel this stuff happened to me also, which was because of the '@typescript-eslint/no-unnecessary-condition': 'warn' rule, which I switched to 'off' in my PR, because it was actually removing question marks which lead to breaking changes.

interfaces/IBF-dashboard/src/app/services/map.service.spec.ts
```diff
@@ -20,9 +29,57 @@ describe('MapService', () => {
...
+  describe('getAreaByPlaceCode', () => {
+    it('should return the triggered area corresponding to the provided placeCode or placeCodeParent', () => {
+      service.triggeredAreas = MOCK_TRIGGEREDAREAS;
+
+      expect(
+        service.getAreaByPlaceCode(
+          MOCK_PLACECODE.placeCode,
+          MOCK_PLACECODE.placeCodeParent.placeCode,
+        ),
+      ).toEqual(MOCK_TRIGGEREDAREAS[0]);
```
jannisvisser, line 82: It doesn't feel intuitive why this is expected to equal. Maybe add 'const expectedArea = MOCK_TRIGGEREDAREAS[0]', or something else to make it more clear.

Also, I feel you're mixing up the use of MOCK_PLACECODE a bit. I think mock in this context is really meant to be used for defining responses (or intermediate data, like service.triggeredAreas), and not for input. You don't need to mock the input right? You can simply pass the input.. So it's mostly semantics, but I feel like it's important to split.. So just say 'const placeCode = abc' and pass that in the method.

interfaces/IBF-dashboard/src/app/services/map.service.spec.ts
```diff
@@ -20,9 +29,57 @@ describe('MapService', () => {
...
+    });
+
+    it('should be undefined if no layer has the provided name', () => {
+      const layerName = IbfLayerName.gauges;
+
+      expect(service.getLayerByName(layerName)).toBeUndefined();
+    });
+  });
+
+  describe('getPlaceCodeParent', () => {
```
jannisvisser, line 59: can you comment on why you chose the methods you chose?

jannisvisser (changes requested):
Looks good Andrea. Maybe read through the comments and then shortly call about it?

interfaces/IBF-dashboard/src/app/services/map.service.spec.ts
```diff
@@ -22,7 +22,7 @@ describe('MapService', () => {
 
   beforeEach(() => {
     TestBed.configureTestingModule({
-      imports: [RouterTestingModule, TranslateModule.forRoot()],
+      imports: [RouterModule.forRoot([]), TranslateModule.forRoot()],
```
jannisvisser, line 25: @arsforza please replace RouterTestingModule throughout all current spec files

interfaces/IBF-dashboard/src/app/services/map.service.spec.ts
```diff
@@ -1,28 +1,102 @@
...
+    it('should be undefined if no layer has the provided name', () => {
+      const layerName = IbfLayerName.gauges;
+
+      expect(service.getLayerByName(layerName)).toBeUndefined();
+    });
+  });
+
+  describe('getPlaceCodeParent', () => {
+    it('should return the placeCodeParent of the provided placeCode', () => {
+      // eslint-disable-next-line @typescript-eslint/dot-notation
```
jannisvisser, line 61: is it easy to process the rules correctly here? and if not, is it then best practice to use this individual ignores, instead of just setting the rule to 'off' for now?

interfaces/IBF-dashboard/src/app/services/map.service.spec.ts
```diff
@@ -1,28 +1,102 @@
...
+      adminLevelService['countryDisasterSettings'] =
+        MOCK_COUNTRYDISASTERSETTINGS;
+      // eslint-disable-next-line @typescript-eslint/dot-notation
+      adminLevelService['adminLevel'] = 3;
+
+      const placeCode: PlaceCode = {
```
jannisvisser, line 67: given that you use this also in the getAreaByPlaceCode test, maybe you can define this at the top of the file, and then in the getAreaByPlaceCode test say 'const placeCode = placeCode.placeCode' etc.

jannisvisser (changes requested):
small changes, for the rest looks good now I think!

jannisvisser (approved):

Issue #1602, opened 2024-09-30
Remove nginx from docker compose


gulfaraz:
### **Who is your contact person for this task?**
@gulfaraz 

Nginx was added to the IBF setup because the _vision_ was that there would be many IBF instances. However, this is not the case anymore. We should move the nginx config outside the docker-compose setup and make it a part of the host machine (similar to the webhook). For development, we can keep a nginx image in the override file without the letsencrypt and certbot dependencies.

```[tasklist]
### **What needs to be done**
- [ ] Update the docker compose files
- [ ] Update any relevant documentation
- [ ] Setup Nginx on host machine in prod, demo, and test
```

```[tasklist]
### **Acceptance Criteria**
- [ ] Tested and reviewed
```

### **Additional context**
[nginx folder in the repo](https://github.com/rodekruis/IBF-system/tree/master/nginx)
https://github.com/rodekruis/IBF-system/blob/5bb38d45f41e84c4e890b04315f7593b8bdb8be7/docker-compose.yml#L2-L17

gulfaraz:
PR #1815 is merged into master after review with @jannisvisser

PR #1733, opened 2024-10-29
build: remove nginx from docker compose

gulfaraz:
## To do
- [x] Get ibf-dashboard setup working locally
- [x] Get ibf-dashboard setup working on ibf-test
- [ ] Get connection to api-service/geoserver working
- [ ] secure access
- [ ] Clean up and update VM_SETUP.md
- [ ] Review
- [ ] Merge + release + setup on ibf-demo
- [ ] Setup on ibf-prod

### Goal based on issue [description](https://github.com/rodekruis/IBF-system/issues/1602#issue-2557670046)
The goal is to remove the dependency of nginx in IBF deployments. This ties in with the Orca alerts as the nginx image is flagged as a concern. From a devops perspective, this change will allow IBF to be setup behind other web servers like Apache. However, the config we provide support for is nginx because that is what we use for our own deployments.

### Plan on local to commit
1. Remove nginx image from main docker compose file. ✅
2. Add nginx image to docker compose override for use in dev setup. ✅
3. Move the nginx conf into the tools folder for our own deployments. ✅
    1. We will (manually) use this conf when we setup the VM. ✅
    2. We will use this conf for dev setup. ✅
4. Expose the necessary ports in docker compose.
    1. Before, the web server routing was within the docker network so the ports did not need to be exposed.
    2. Now, nginx will run on the VM and it will route incoming requests to the respective ports.
    3. The urls should be changed in the nginx conf file accordingly at the time of setup. I was planning to have a commented line http://localhost:3000 below http://ibf-api-service:3000. Then we can (manually) swap this line when we setup the VM for deploy.
5. Update the VM setup docs to include instructions for the manual nginx VM setup.

### Plan on VMs
1. Update the test env with the new config.
    1. Deploy the new version.
    2. Install/enable nginx on the VM.
    3. Update the nginx conf and reload nginx.
    4. Install and run certbot on the VM using certbot --nginx -d ibf-test.510.global. https://certbot.eff.org/instructions?ws=nginx&os=ubuntufocal
    5. Check whether the web server routes properly. The dashboard should load, the api are responsive directly and via the dashboard, geoserver loads via the dashboard.
2. If needed, revise the nginx [VM setup docs](https://github.com/rodekruis/IBF-system/blob/master/docs/VM_SETUP.md).
3. Follow nginx VM setup doc on demo.
4. If needed, revise the nginx VM setup docs.
5. Follow nginx VM setup docs on prod.

### Scenarios to consider
1. The new setup should create the required images. When the containers run, the dashboard should be present on the host VM.
2. Containers are stopped and restarted.
3. Images are removed and rebuilt.
4. CI setup should work. This can be checked by manually running deploy.sh.
    - I noticed that when node_modules and www are present but have no content. Then they don’t get filled with the required files. I believe but unsure if this can be addressed by removing the folders in the deploy.sh script after the docker compose down.
5. Check if the setup works when a node module is changed. I want to confirm if the node modules get updated if we change something in package.json.
6. Manually delete the contents of node modules and the www folders to check if they get rebuilt.

I rely on adjusting the above based on what I encounter during execution. So take the above as a general direction and update accordingly. Let me know if you have questions @jannisvisser 

---

## Describe your changes

## Notes for the reviewer

### DIFF
--- a/docker-compose.ci.yml
+++ b/docker-compose.ci.yml
@@ -16,9 +16,9 @@ services:
   ibf-dashboard:
     entrypoint: ['echo', 'Service ibf-dashboard disabled']
 
-  ibf-geoserver:
-    ports:
-      - 8081:8080
+  # ibf-geoserver:
+  #   ports:
+  #     - 8081:8080
 
   ibf-local-db:
     image: postgis/postgis
--- a/docker-compose.ci.yml
+++ b/docker-compose.ci.yml
@@ -10,7 +10,6 @@ services:
     depends_on:
       - ibf-local-db
     networks:
-      - web-server-network
       - api-network
     user: root
 
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -26,9 +26,9 @@ services:
   ibf-dashboard:
     entrypoint: ['echo', 'ibf-dashboard is disabled']
 
-  ibf-geoserver:
-    ports:
-      - 8081:8080
+  # ibf-geoserver:
+  #   ports:
+  #     - 8081:8080
 
   ibf-local-db:
     image: postgis/postgis
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -1,4 +1,15 @@
 services:
+  nginx:
+    image: nginx
+    ports:
+      - 8080:80
+    volumes:
+      - ./tools/nginx.conf:/etc/nginx/conf.d/default.conf
+      - ./interfaces/IBF-dashboard/www/browser:/var/www/ibf-dashboard
+    networks:
+      - api-network
+    restart: unless-stopped
+
   ibf-api-service:
     command: ['npm', 'run', 'start:dev']
     environment:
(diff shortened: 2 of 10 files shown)


jannisvisser:
@gulfaraz general overview is clear, but I guess I will run into the details when I actually dive into it. So be it.

gulfaraz:
Nice, thank you for reviewing the handover.

jannisvisser:
Status
- I cannot get this to work yet, and am hesitant to spend too much time on it as I feel things might be much faster when being able to spar about it together. So given there's no immediate hurry, I might wait until Gulfaraz is back.
- Specifically, a main missing point is that we have to copy the built version of ibf-dashboard to /var/www/ibf-dashboard (or something similar), but I'm not getting any output in my /www folder even though there is a volume. 
- I have tried developing/testing this by checking out this branch on ibf-test. Reverse this through:
  - git reset --hard origin/master
  - sudo service nginx stop
  - . ./tools/deploy.sh

gulfaraz:
Closing this PR in favour of #1815

PR #1775, opened 2024-11-11
Add tests for the alert threshold feature

Piotrk39:
## Describe your changes

Adds E2E tests for alert thresholds lines both Trigger and no Trigger mode.

## Notes for Reviewer

The test for testing the Threshold alert on the legend is failing because on No trigger mode the element is visible on the legend and it shouldn't. Therefore the decision has to be made if the test should be skipped.


tests/e2e/Pages/MapComponent.ts
```diff
@@ -188,6 +192,29 @@ class MapComponent extends DashboardPage {
     await expect(legendComponent).toBeVisible();
   }
 
+  async assertLegendElementIsNotVisible({
```
jannisvisser, line 195: this new method is not actually used anywhere. Instead the version without 'Not' in the name is used, which is probably better any way for intuitive interpretation.

tests/e2e/Pages/MapComponent.ts
```diff
@@ -46,6 +47,9 @@ class MapComponent extends DashboardPage {
     this.layerMenuToggle = this.page.getByTestId('layer-menu-toggle-button');
     this.redCrossMarker = this.page.getByAltText('red-cross-branch-marker');
     this.gloFASMarker = this.page.getByAltText('glofas-station-marker');
+    this.alerThresholdLines = this.page.locator(
+      '[stroke="var(--ion-color-ibf-outline-red)"]',
```
jannisvisser, line 51: is this safe enough? What is 'stroke' here exactly? Can there be nothing else that ends up in this filter?

jannisvisser (changes requested):

tests/e2e/tests/Map/NoAlertThresholdIsVisibleNoTrigger.spec.ts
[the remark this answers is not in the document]
Piotrk39: I believe that since we have rather vague documentation so far it is good to keep them separate in case we change something. Then only one of those tests will fail. I change the file name.

tests/e2e/Pages/MapComponent.ts
Piotrk39: This is not ideal but after careful consideration this was the perfect and only candidate. That is why I kept it. It is also safe enough since it only appears on this particular element. Time will tell with other tests if it has to be updated. For now to save the dev time it is the only option.

tests/e2e/Pages/MapComponent.ts
Piotrk39: I removed it. It was created with different scope in mind but since we check for element to be visible I had to replace it with the old function. Thanks!

jannisvisser (approved):

PR #1801, opened 2024-11-25
Add e2e test for info icons interactions

Piotrk39:
## Describe your changes

Adds e2e test for checking interactions with map layer info icons

### DIFF
--- a/tests/e2e/Pages/MapComponent.ts
+++ b/tests/e2e/Pages/MapComponent.ts
@@ -21,6 +21,8 @@ class MapComponent extends DashboardPage {
   readonly redCrossMarker: Locator;
   readonly gloFASMarker: Locator;
   readonly alerThresholdLines: Locator;
+  readonly closeButtonIcon: Locator;
+  readonly layerInfoContent: Locator;
 
   constructor(page: Page) {
     super(page);
@@ -50,6 +52,8 @@ class MapComponent extends DashboardPage {
     this.alerThresholdLines = this.page.locator(
       '[stroke="var(--ion-color-ibf-outline-red)"]',
     );
+    this.closeButtonIcon = this.page.getByTestId('close-matrix-icon');
+    this.layerInfoContent = this.page.getByTestId('layer-info-content');
   }
 
   async mapComponentIsVisible() {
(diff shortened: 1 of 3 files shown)


tests/e2e/Pages/MapComponent.ts
```diff
@@ -194,6 +207,39 @@ class MapComponent extends DashboardPage {
     return availableLayers;
   }
 
+  async assertAllInfoIconsIntercations() {
```
jannisvisser, line 210: typo (Intercations)

tests/e2e/tests/Map/InfoIconsAreInteractiveInLayerComponent.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  await page.goto('/');
+  await loginPage.login(
+    NoTriggerDataSet.UserMail,
+    NoTriggerDataSet.UserPassword,
+  );
+});
+
+test(
+  qase(33, 'Interaction with info icon in map layer component'),
```
jannisvisser, line 39: is it possible and/or interesting and/or feasible to add a link to the qase test case behind the number 33 here (and for all other test cases?). Feel free to say 'good idea, but too much work for now'.

tests/e2e/tests/Map/InfoIconsAreInteractiveInLayerComponent.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    // Wait for the page to load
+    await dashboard.waitForLoaderToDisappear();
+    await map.mapComponentIsVisible();
+
+    // Open the layer menu
+    await map.isLayerMenuOpen({ layerMenuOpen: false });
+    await map.clickLayerMenu();
+    await map.isLayerMenuOpen({ layerMenuOpen: true });
+
+    // Assert layer info icons to be intercative and contain basic required info
```
jannisvisser, line 60: what basic required info is asserted then actually? just the 'trimmedName', which is basically that the layer name also appears in the popup again?

jannisvisser (changes requested):
looks good! very minor comments/questions.

tests/e2e/tests/Map/InfoIconsAreInteractiveInLayerComponent.spec.ts
Piotrk39: It is a name of the layer the info button was clicked for as it should appear in the header of the pop-up as well. And then the content has just to be present. Unfortunately each info content is different not all of them have links and some are missing. Do you think we should investigate it to keep them all in the same required pattern?

jannisvisser: no it's fine, just wanted to check

tests/e2e/tests/Map/InfoIconsAreInteractiveInLayerComponent.spec.ts
jannisvisser: I guess it would make sense to add them for all test cases so far I guess, but agree on switching to Azure so good enough for now. But if you feel like it in any of the upcoming items still, go ahead.

jannisvisser (approved):

PR #1830, opened 2024-12-03
fix: api test flash floods AB#31853

jannisvisser:
## Describe your changes

AB#31853

## Notes for the reviewer

1. Feel free to have an opinion on this taking so long (because the mock takes long). It's quite annoying both for manual mocking and for API-test that takes long. So should we spend effort on making the mock quicker? Or change api-test setup (especially the new blantyre city mock is slow, so maybe add a new scenario without blantyre, which is used for api test?). 
2. This adds to a context where the time all the API-tests take is already accumulating fast. Can we look at this? Is the reset+mock needed for every test again (I think so..)


jannisvisser:
@gulfaraz I opted for a shorter scenario for the api test

gulfaraz (commented):
The term `api-test` doesn't describe the scenario. Going by the description let's use `kargona-rumphi-trigger` instead of `api-test` and `blantyre-trigger` instead of `single-event`.

The increased timeout needs to be addressed. I see the long delay on my local. Let's understand why this happens before we merge the workaround.

jannisvisser:
@gulfaraz currently master branch has a failing test, so I would suggest to merge this quickly actually. We can look at the api-test speed in general in the next api-test item (I've made a task already).

Also, the scenario-names are theoretically not country-specific, as they come from a flash floods enum. So by keeping the names more generic there's flexibility to fill them per country differently. Of course there's only 1 country for flash floods now, but still.

gulfaraz:
> currently master branch has a failing test, so I would suggest to merge this quickly actually. We can look at the api-test speed in general in the next api-test item (I've made a task already).

Let's merge it first then fix it after.

> Also, the scenario-names are theoretically not country-specific, as they come from a flash floods enum. So by keeping the names more generic there's flexibility to fill them per country differently. Of course there's only 1 country for flash floods now, but still.

Good point. I misunderstood this because although the scenario names are generic, their description is not.

https://github.com/rodekruis/IBF-system/blob/b55a7f5f6eebf104da83e1aaa1c2d8645daac000/services/API-service/src/scripts/mock.controller.ts#L75

`single-event` and `api-test` are still unclear to me. However I agree the name and description be generic,
- `one-area-trigger` and `two-areas-trigger`
- `one-trigger` and `two-triggers`
- `single-event` and `double-event`
- `blantyre-trigger` and `karonga-rumphi-trigger` (works only until we add flash-flood to another country)

The [API description](https://github.com/rodekruis/IBF-system/blob/b55a7f5f6eebf104da83e1aaa1c2d8645daac000/services/API-service/src/scripts/mock.controller.ts#L75) should also be updated based on which name we choose.

gulfaraz:
@jannisvisser makes sense. Do you have a preference for any option? If they're all the same to you then I will pick the last one as it's the easiest to interpret.

PR #1832, opened 2024-12-04
Add e2e test for default flood trigger layers

Piotrk39:
## Describe your changes

Adds test for default triggered layers Floods

### DIFF
--- a/tests/e2e/tests/Map/DefaultLegendLayersInTriggeredFlood.spec.ts
+++ b/tests/e2e/tests/Map/DefaultLegendLayersInTriggeredFlood.spec.ts
@@ -1,4 +1,4 @@
-import { expect, test } from '@playwright/test';
+import { test } from '@playwright/test';
 import DashboardPage from 'Pages/DashboardPage';
 import MapComponent from 'Pages/MapComponent';
 import UserStateComponent from 'Pages/UserStateComponent';
@@ -53,7 +53,11 @@ test(
 
     await map.mapComponentIsVisible();
     await map.isLegendOpen({ legendOpen: true });
-    const legendLayers = await map.returnLayerCheckedCheckboxes();
-    expect(legendLayers).toEqual(['Flood extent', 'Exposed population']);
+    await map.assertLegendElementIsVisible({
+      legendComponentName: 'Flood extent',
+    });
+    await map.assertLegendElementIsVisible({
+      legendComponentName: 'Exposed population',
+    });
   },
 );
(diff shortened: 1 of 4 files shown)


tests/e2e/tests/Map/DefaultLegendLayersInTriggeredFlood.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+  );
+
+  await page.goto('/');
+  await loginPage.login(TriggerDataSet.UserMail, TriggerDataSet.UserPassword);
+});
+// https://app.qase.io/project/IBF?previewMode=side&suite=3&tab=&case=37
+test(
+  qase(
+    37,
+    '[Trigger] Map layer: "Flood extent" and "Exposed population" should be active by default',
```
jannisvisser, line 38: What is the exact purpose of this test. I often get mixed up by filenames vs test descriptions here vs full test cases in Qase. Is in this case the test just about the layers being present in the legend? And with legend do you mean the lower-left thing or the lower-right thing? 

Either way, I think it makes sense to test all together. So test (1) are the layers in the layer overview (lower-right), (2) are they in the legend (lower-left), and (3) are they in the map? I'm not clear on if this is happening now either in this test or in multiple tests.

jannisvisser (changes requested):

tests/e2e/tests/Map/DefaultLegendLayersInTriggeredFlood.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    // Assertions
+    await userState.headerComponentIsVisible({
+      countryName: TriggerDataSet.CountryName,
+    });
+    // Wait for the page to load
+    await dashboard.waitForLoaderToDisappear();
+
+    await map.mapComponentIsVisible();
+    await map.isLegendOpen({ legendOpen: true });
+    await map.assertLegendElementIsVisible({
```
jannisvisser, line 56: now only legend is tested. test also layer overview and map (proxy via radio-btn being active)

tests/e2e/tests/Map/ExposedPopulationVisibleInTrigger.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+    accessToken,
+  );
+
+  await page.goto('/');
+  await loginPage.login(TriggerDataSet.UserMail, TriggerDataSet.UserPassword);
+});
+
+test(
+  qase(
+    36,
```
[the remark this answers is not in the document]
jannisvisser: This test case maybe overlaps iwth 37, drop this one?

PR #1834, opened 2024-12-06
refactor: fix integration test flash flood

gulfaraz:
## Describe your changes

1. Rename the scenarios as discussed [here](https://github.com/rodekruis/IBF-system/pull/1830#issuecomment-2522534964).
2. Configure `isApiTest` to be`false` by default instead of `undefined`.
3. Optimize [uploadAssetExposureStatus in LinesDataService](https://github.com/rodekruis/IBF-system/blob/ef01f8b8ffdf75795af2d0b7f70821f3f3173fae/services/API-service/src/api/lines-data/lines-data.service.ts#L115) to reduce db queries.

## Notes for the reviewer

The mock endpoint is slow because of [the for-loop in LinesDataService](https://github.com/rodekruis/IBF-system/blob/f11fc1c60c7740d3467d26a48ffeb5e88c6a9cc2/services/API-service/src/api/lines-data/lines-data.service.ts#L121). It is faster if we use a single database query with a list of ids instead of a separate db query for each id.

> ...it's usually better to perform one database query to get all the data you need and then start operating on them than to query, process, query again, process again, etc. It dramatically decreases the overall time needed to finish the loop. - [SO](https://softwareengineering.stackexchange.com/a/367829/312865)


tests/integration/email/flash-flood/email-mwi-flash-flood.test.ts
```diff
@@ -11,16 +11,49 @@ describe('Should send an email for mwi flash flood', () => {
     await resetDB(accessToken);
   });
 
-  it('api-test', async () => {
-    const eventNames = ['Karonga']; // Scenario contains also 'Rumphi' but as ongoing, for which no email is sent
+  it('trigger', async () => {
+    const eventNames = ['Blantyre City', 'Karonga'];
```
jannisvisser, line 15:
```suggestion
    const eventNames = ['Blantyre City', 'Karonga'];  // Scenario contains also 'Rumphi' but as ongoing, for which no email is sent
```

jannisvisser (approved):

PR #1854, opened 2024-12-16
chore: set up husky & lint-staged AB#32125

jannisvisser:
## Describe your changes

Copy husky & lint-staged setup from 121, to make sure we properly run prettier and lint automatic fixes on commit, and only on staged files.


arsforza (approved):

services/API-service/package.json
```diff
@@ -78,5 +78,13 @@
     "tsconfig-paths": "^3.9.0",
     "xlsx": "https://cdn.sheetjs.com/xlsx-0.20.3/xlsx-0.20.3.tgz",
     "yargs": "^17.0.0"
+  },
+  "lint-staged": {
+    "**/*.{js,jsx,ts,tsx,html,css,scss,json,md}": [
```
aberonni, line 83: You are deviating from what is laid out in the root directory here (eg. using prettier instead of eslint for js files), but maybe this is an intentional override.

aberonni (commented):
For the most part looks good to me. I also found that we had to iterate a few times on the 121-platform until we ironed out all the kinks, so it might be ok to just get people using it and then adjust accordingly.

services/API-service/package.json
jannisvisser: it was not, I'll adjust, thanks!

.husky/pre-commit
```diff
@@ -0,0 +1,27 @@
...
+NC='\033[0m' # No Color
+
+log() {
+  printf "${ORANGE}[pre-commit]${NC} $1\n"
+}
+
+log "Running custom pre-commit hook defined in ./husky/pre-commit"
+
+# Prevent committing to "main"-branch:
+if [ "$(git rev-parse --abbrev-ref HEAD)" = "main" ]; then
```
gulfaraz, line 16: @jannisvisser @arsforza This will not work because IBF does not have a [`main`](https://github.com/rodekruis/IBF-system/tree/main) branch. It does have a [`master`](https://github.com/rodekruis/IBF-system/tree/master) branch which we could use. Or create a `main` branch and use that instead of `master`.

jannisvisser: good point I'll correct it

PR #1867, opened 2024-12-19
Fix.toggle off glofas layer

jannisvisser:
## Describe your changes

Resolves [AB#31737](https://dev.azure.com/redcrossnl/9e7ca3cc-bb97-4471-a25c-fd9787af0fe8/_workitems/edit/31737)

## Notes for the reviewer

### DIFF
--- a/interfaces/IBF-dashboard/src/app/components/matrix/matrix.component.html
+++ b/interfaces/IBF-dashboard/src/app/components/matrix/matrix.component.html
@@ -24,7 +24,7 @@
           <ion-row class="ion-justify-content-between ion-nowrap">
             <ion-item data-testid="matrix-layer-name" lines="none">
               @if (layer.isLoading) {
-                <ion-spinner></ion-spinner>
+                <ion-spinner class="mx-[0.15rem]"></ion-spinner>
               } @else {
                 @if (isCheckBox(layer.group)) {
                   <ion-icon
--- a/interfaces/IBF-dashboard/src/app/services/map.service.ts
+++ b/interfaces/IBF-dashboard/src/app/services/map.service.ts
@@ -300,18 +300,12 @@ export class MapService {
   };
 
   private loadWaterPointsLayer = (layer: IbfLayerMetadata) => {
-    const layerDataCacheKey = this.getLayerDataCacheKey(layer.name);
-    if (this.layerDataCache[layerDataCacheKey]) {
-      this.addWaterPointsLayer(layer, this.layerDataCache[layerDataCacheKey]);
-    } else {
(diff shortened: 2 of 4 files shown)


gulfaraz:
https://github.com/rodekruis/IBF-system/blob/5956067133c0aeccf4fdcae1554cf0e562ac91be/interfaces/IBF-dashboard/src/app/services/map.service.ts#L610

This check stops loading data for inactive layers. This check is made in the [`updateLayers`](https://github.com/rodekruis/IBF-system/blob/5956067133c0aeccf4fdcae1554cf0e562ac91be/interfaces/IBF-dashboard/src/app/services/map.service.ts#L596C11-L596C23) function which is currently only used in [`toggleLayer`](https://github.com/rodekruis/IBF-system/blob/5956067133c0aeccf4fdcae1554cf0e562ac91be/interfaces/IBF-dashboard/src/app/services/map.service.ts#L582). [`toggleLayer`](https://github.com/rodekruis/IBF-system/blob/5956067133c0aeccf4fdcae1554cf0e562ac91be/interfaces/IBF-dashboard/src/app/services/map.service.ts#L582) is called only after user interaction so this check is applied only on active layers. In case we need [`updateLayers`](https://github.com/rodekruis/IBF-system/blob/5956067133c0aeccf4fdcae1554cf0e562ac91be/interfaces/IBF-dashboard/src/app/services/map.service.ts#L596C11-L596C23) to be callable from other functions we want to keep the active layer restriction.

The GloFAS layer's active flag is properly set. The reason it does not get updated is because the function does not reach this block,
https://github.com/rodekruis/IBF-system/blob/5956067133c0aeccf4fdcae1554cf0e562ac91be/interfaces/IBF-dashboard/src/app/services/map.service.ts#L606-L610
If the layer is active and loaded, the function assumes that the layer data is also cached. However, the layer data is not cached. The only caching I could find is for the waterpoints layer. I assume we want to cache all layer data if possible. So I enabled layer data caching with these lines,
https://github.com/rodekruis/IBF-system/blob/5956067133c0aeccf4fdcae1554cf0e562ac91be/interfaces/IBF-dashboard/src/app/services/map.service.ts#L481-L487

Then I removed the now redundant waterpoints cache. I also aligned the spacing of the loading spinner coz it was off by a bit.

gulfaraz:
The layer data stopped caching from https://github.com/rodekruis/IBF-system/commit/8a13c079b10bf27292319fe0c62fdeee553a1636. Specifically when this line was removed,
https://github.com/rodekruis/IBF-system/blob/f0be2fd50d34903819cf66ee4887a667d8370e74/interfaces/IBF-dashboard/src/app/services/map.service.ts#L724

PR #1878, opened 2024-12-30
Add E2E test for: [Trigger] Info icon is clickable and opens popover

Piotrk39:
## Describe your changes

Adds test for validating info icons by warnings in chat section.

### DIFF
--- a/tests/e2e/Pages/ChatComponent.ts
+++ b/tests/e2e/Pages/ChatComponent.ts
@@ -26,6 +26,7 @@ class ChatComponent extends DashboardPage {
   readonly windowsOsLink: Locator;
   readonly macOsLink: Locator;
   readonly linuxOsLink: Locator;
+  readonly infoButtons: Locator;
 
   constructor(page: Page) {
     super(page);
@@ -48,6 +49,7 @@ class ChatComponent extends DashboardPage {
     this.windowsOsLink = this.page.getByTestId('export-view-windows-os');
     this.macOsLink = this.page.getByTestId('export-view-mac-os');
     this.linuxOsLink = this.page.getByTestId('export-view-linux-os');
+    this.infoButtons = this.page.getByTestId('tooltip-button');
   }
 
   async chatColumnIsVisibleForNoTriggerState({
@@ -193,6 +195,16 @@ class ChatComponent extends DashboardPage {
       await expect(predictionButton).toBeEnabled();
     }
     return countPredictionButtons;
+  }
+
+  async validateInfoButtonsAreClickable() {
+    const countInfoButtons = await this.infoButtons.count();
+    for (let i = 0; i < countInfoButtons; i++) {
+      const infoButton = this.infoButtons.nth(i);
+      await infoButton.click();
+      await infoButton.click();
+    }
+    return countInfoButtons;
   }
 }
 
--- a/tests/e2e/Pages/ChatComponent.ts
+++ b/tests/e2e/Pages/ChatComponent.ts
@@ -28,7 +28,7 @@ class ChatComponent extends DashboardPage {
   readonly windowsOsLink: Locator;
   readonly macOsLink: Locator;
   readonly linuxOsLink: Locator;
-  readonly infoButtons: Locator;
+  readonly tooltipButton: Locator;
   readonly backDrop: Locator;
 
   constructor(page: Page) {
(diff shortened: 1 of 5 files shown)


tests/e2e/tests/ChatSection/ChatInfoButtonOpenPopOver.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+  await mockFloods(
+    TriggerDataSet.TriggerScenario,
+    TriggerDataSet.CountryCode,
+    accessToken,
+  );
+
+  await page.goto('/');
+  await loginPage.login(TriggerDataSet.UserMail, TriggerDataSet.UserPassword);
+});
+// https://app.qase.io/project/IBF?case=45&previewMode=side&suite=6
```
jannisvisser, line 33: the test case in Qase is not filled in. I think it makes sense to always start with that right before writing the test?

tests/e2e/Pages/ChatComponent.ts
```diff
@@ -194,6 +200,17 @@ class ChatComponent extends DashboardPage {
     }
     return countPredictionButtons;
   }
+
+  async validateInfoButtonsAreClickable() {
+    const countInfoButtons = await this.infoButtons.count();
```
jannisvisser, line 205: so this tests all occurences of <app-tooltip> in the portal? so both the one in chat.component.html and in event-speech-bubble.component? (I guess not the one in river-gauge-popup-content as that is not opened at this stage).

Is it indeed the purpose of this test to test both the overall warning message tooltip, and the event-specific tooltip (see screenshot)

tests/e2e/tests/ChatSection/ChatInfoButtonOpenPopOver.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    const userState = new UserStateComponent(page);
+    const chat = new ChatComponent(page);
+
+    // Navigate to disaster type the data was mocked for
+    await dashboard.navigateToFloodDisasterType();
+    // Assertions
+    await userState.headerComponentIsVisible({
+      countryName: TriggerDataSet.CountryName,
+    });
+    await chat.chatColumnIsVisibleForTriggerState({
```
jannisvisser, line 47: is this meant as a check to wait for the chat section to be fully loaded? if so, does it thereby fully cover another test case already maybe, which could therefore be deleted?

jannisvisser (changes requested):
he piotr, test itself looks clear I think, but still have some questions.

tests/e2e/tests/ChatSection/ChatInfoButtonOpenPopOver.spec.ts
Piotrk39: I forgot to paste, it should be there now.

tests/e2e/Pages/ChatComponent.ts
Piotrk39: It is supposed to test only the info buttons by the events.

tests/e2e/tests/ChatSection/ChatInfoButtonOpenPopOver.spec.ts
Piotrk39: I will actually come back to it when I do the refactor because many tests theoretically can be combined. I will do it soon in a separate PR. If this is alright with you?

tests/e2e/Pages/ChatComponent.ts
jannisvisser: OK, cleared up that the top info-btn is only showing when the text is red, so not appearing in the test case.

Agreed to try to instead put the testId in app-tooltip of event-speech-bubble.component instead of tooltip.component itself.

Piotrk39: It seems like the tooltips were separated between trigger event and event so I added it twice. Please verify that this is indeed correct implementation.

jannisvisser: I think that's fine. But I'm surprised the test succeeds, because this.infoButtons is now an array of <app-tooltips> and no longer an array of <ion-buttons>, so in line 209 below I would think that infoButton.click() shouldn't work any more. And instead you would need to locate the ion-button element. Can you confirm this is going allright? Also, the naming should be changed accordingly.

Piotrk39: In practice it does not matter how the array is constructed as long as the testid is the same for both buttons they are treated equally. I can of course change it for event-tooltip and trigger-event-tooltip if it makes it clearer. I also changed the name of the function and it's variables.

jannisvisser (approved):

PR #1900, opened 2025-01-06
test: unit tests for country module

gulfaraz:
## Describe your changes

[AB#31781](https://dev.azure.com/redcrossnl/9e7ca3cc-bb97-4471-a25c-fd9787af0fe8/_workitems/edit/31781)

## Notes for the reviewer

1. I did not mock private functions based on [dev chat](https://teams.microsoft.com/l/message/19:bdb9d621f18c48c7917e545bf6d7aabe@thread.v2/1736170063678?context=%7B%22contextType%22%3A%22chat%22%7D).
2. No single expectation of a function exists. The functions are implemented to perform multiple data changes so I added an assert for each.
3. I skipped `getBoundingBoxWkt` because it is a single TypeORM statement.

### DIFF
--- a/services/API-service/src/api/country/country.service.spec.ts
+++ b/services/API-service/src/api/country/country.service.spec.ts
@@ -182,9 +182,9 @@ describe('CountryService', () => {
           relations: ['notificationInfo'],
         });
       }
-      expect(countryRepository.save).toHaveBeenCalledWith(
-        notificationInfos.map(() => ({ notificationInfo: null })),
-      );
+      expect(countryRepository.save).toHaveBeenCalledWith([
+        { notificationInfo: {} },
+      ]);
     });
   });
 
--- a/services/API-service/src/api/country/country.service.spec.ts
+++ b/services/API-service/src/api/country/country.service.spec.ts
@@ -10,10 +10,14 @@ import { NotificationInfoEntity } from '../notification/notifcation-info.entity'
 import { CountryDisasterSettingsEntity } from './country-disaster.entity';
 import { CountryEntity } from './country.entity';
 import { CountryService } from './country.service';
+import { CountryDisasterSettingsDto } from './dto/add-countries.dto';
 
 describe('CountryService', () => {
   let service: CountryService;
   let countryRepository: Repository<CountryEntity>;
+  let disasterRepository: Repository<DisasterEntity>;
+  let countryDisasterSettingsRepository: Repository<CountryDisasterSettingsEntity>;
+  let leadTimeRepository: Repository<LeadTimeEntity>;
 
   const relations = [
     'countryDisasterSettings',
@@ -52,6 +56,15 @@ describe('CountryService', () => {
     service = module.get<CountryService>(CountryService);
     countryRepository = module.get<Repository<CountryEntity>>(
       getRepositoryToken(CountryEntity),
+    );
+    disasterRepository = module.get<Repository<DisasterEntity>>(
+      getRepositoryToken(DisasterEntity),
+    );
+    countryDisasterSettingsRepository = module.get<
+      Repository<CountryDisasterSettingsEntity>
+    >(getRepositoryToken(CountryDisasterSettingsEntity));
+    leadTimeRepository = module.get<Repository<LeadTimeEntity>>(
+      getRepositoryToken(LeadTimeEntity),
     );
   });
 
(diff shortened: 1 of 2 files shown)


services/API-service/src/api/country/country.service.spec.ts
```diff
@@ -0,0 +1,293 @@
...
+
+      expect(await service.getCountries(countryCodes)).toBe(result);
+      expect(countryRepository.find).toHaveBeenCalledWith({
+        where: { countryCodeISO3: In(countryCodes.split(',')) },
+        relations,
+      });
+    });
+  });
+
+  describe('findOne', () => {
```
jannisvisser, line 105: I see this findOne method is actually only public at the moment for the sake of the unit test, which is a mistake. It should have been private, and therefore not be unit tested. It is only used by the method addOrUpdateNotificationInfo , so we can easily delete it. If you want to preserve the 2 different scenarios you now have in findOne, you could move those to the addOrUpdateNotificationInfo  test?

services/API-service/src/api/country/country.service.spec.ts
```diff
@@ -0,0 +1,293 @@
...
+      );
+      expect(countryRepository.findOne).toHaveBeenCalledWith({
+        where: { countryCodeISO3 },
+        relations: customRelations,
+      });
+    });
+  });
+
+  describe('addOrUpdateCountries', () => {
+    it('should add or update countries', async () => {
```
jannisvisser, line 135: this is quite a long test. I think it helps for readability already to add the //Arrange, //Act and //Assert steps in the code.

services/API-service/src/api/country/country.service.spec.ts
```diff
@@ -0,0 +1,293 @@
...
+        );
+      }
+      expect(countryRepository.save).toHaveBeenCalledWith(
+        notificationInfos.map(() => ({ notificationInfo: null })),
+      );
+    });
+  });
+
+  describe('createNotificationInfo', () => {
+    it('should create notification info', async () => {
```
jannisvisser, line 239: can you reflect on 
- the value vs effort of this unit test (other than "in an ideal world we want full coverage with  all public methods unit tested").
- the type of test for this test case

This is mostly a very straightforward database save instruction, where we mocking the database operations themselves. So what's left to actually test is internal business logic (ifs/paths) in this method, which there isn't much. I feel in this case an integration test on the POST notification info endpoint might make more sense.

jannisvisser (changes requested):
nice work! some questions/comments.

services/API-service/src/api/country/country.service.spec.ts
gulfaraz: > for the sake of the unit test

Did we have unit tests before? I couldn't find any. Anyhow, good catch, I've removed it as you suggested and removed its test.

services/API-service/src/api/country/country.service.spec.ts
gulfaraz: I further refactored the test. Split into //Arrange, //Act and //Assert.

services/API-service/src/api/country/country.service.spec.ts
gulfaraz: ### The value of this unit test

In a well-maintained codebase, most functions are simple to write and to test. Not testing a function because it's simple is good motivation for writing complicated functions. When we fix a bug, [more](https://github.com/rodekruis/IBF-system/blob/5e61128e8e226ae8b03f0e5813aca5189ffaf8d4/services/API-service/src/api/point-data/point-data.service.ts#L93) [often](https://github.com/rodekruis/IBF-system/blob/f0be2fd50d34903819cf66ee4887a667d8370e74/interfaces/IBF-dashboard/src/app/services/map.service.ts#L724) [than](https://github.com/rodekruis/IBF-system/blob/237a2dcf5239bab486e5dc16e17409684e51f5f5/services/API-service/src/api/user/dto/create-user.dto.ts#L38) [not](https://github.com/rodekruis/IBF-system/blob/8e46d4fa0bc46ba214fd914dd1b4d1fcc77a547b/services/API-service/src/roles.guard.ts#L24), the problem is caused by _a single line of code_. If we consistently test an extra line of code, eventually the codebase achieves good test coverage. In such a codebase, we will catch such isolated bugs proactively. If we fail to do so, it's easier to add a second test to prevent reintroducing the bug in future changes.

### The effort of this unit test

I spent most of my energy on `addOrUpdateCountries` then much less on `addOrUpdateNotificationInfo` and almost no time on `createNotificationInfo`. For `createNotificationInfo`, I copy-pasted the implementation as it was that simple.

> in an ideal world we want full coverage with all public methods unit tested

The goal is **_not_** full coverage. The goal is to use the codebase with some confidence. We write tests exactly because it is not an ideal world. In an ideal world, we need no tests because the _code does not always work_.

### Type of test
Integration and unit tests [serve different purposes](https://circleci.com/blog/unit-testing-vs-integration-testing/#:~:text=While%20unit%20tests%20always%20take,works%20in%20an%20integrated%20way.). One type does not diminish the value of the other. For [this task](https://dev.azure.com/redcrossnl/IBF/_workitems/edit/31781), I focused on writing only unit tests. I see why an integration test works for `createNotificationInfo`. We should prioritize the [relevant task](https://dev.azure.com/redcrossnl/IBF/_workitems/edit/31789).

### Why did I write this unit test?
1. I could have stopped at 3 tests as planned but I've already done the hard part of gathering the required context for this part of the app. It takes me _much less energy to add 2 more tests now_ rather than revisit this at some point in the future. Hypothetically, if there were 10 more untested functions in the same file, I may have stopped at 5 (I hope) but we don't know this now.
2. I need the practice to develop a habit of writing good tests. I will improve only if I write more tests. I won't magically write the few critical tests one day doing the hard work beforehand.

Let me know if you prefer that I treat the planned number of tests as a hard limit. In this case, it wasn't a significant additional effort but it will help to know for future test PBIs.

### Refactor the code
When understanding the code, I struggled with the temptation to **refactor** the code. When writing tests, it is stabilizing to assume the code is correctly implemented to a reasonable degree. Otherwise, every test item becomes a refactor item. The first refactor I identified was the `findOne` as you reviewed. A second potential refactor is to combine [save and find](https://github.com/rodekruis/IBF-system/blob/335d873828296c3f27dbef25f23c8283c4aa19a8/services/API-service/src/api/country/country.service.ts#L287-L292). I did not want to fall into that trap this early in my testing effort so I didn't act on these. I hope to refactor code in the test PBIs after I gain confidence in my ability to write tests.

services/API-service/src/api/country/country.service.spec.ts
jannisvisser: Good. I did not mean there were unit tests before, I meant that it was incorrectly public before, which allowed us to write a unit test on it inadvertently. Anyway, resolved now!

services/API-service/src/api/country/country.service.spec.ts
jannisvisser: thanks for this write-up, it makes a lot of sense :) 
- Definitely no need to stick to a hard limit. I very much get and agree with the concept that the 4th and 5th test are increasingly easy to write.
- I do think from now on, it is good to pick up the integration test item and the back-end unit test item of one module together (or even combine into one item), to look at this more holistically.

jannisvisser (approved):

PR #1930, opened 2025-01-22
fix: flaky e2e tests AB#32712

jannisvisser:
## Describe your changes

See AB#32712


Piotrk39:
@jannisvisser It all looks good. I believe that due to how the front-end was designed we cannot skip fixed awaits so waiting for a second actually makes it faster in the long run even tho theory says we should not use it like that. I believe we can merge it.

PR #1933, opened 2025-01-22
Un-skip e2e test

Piotrk39:
## Describe your changes

Fix one e2e test that was failing after the Playwright framework refactor.

## Notes for the reviewer

I have switched the data-testids that were used for selecting the locators. It seems that they were not visible anymore after the refactor even tho there seems to be no connection between the refactor and testids. I have reused old ids by adding playwright readable format to them 'data-testid' instead of 'data-test'. Also I refactored the test itself to use those ids correctly.

### DIFF
--- a/tests/e2e/tests/UserStateComponent/UserStateComponentVisible.ts
+++ b/tests/e2e/tests/UserStateComponent/UserStateComponentVisible.ts
@@ -10,7 +10,7 @@ export default (
   disasterType: string,
 ) => {
   // Skip for now, as test often fails
-  test.skip(
+  test(
     qase(3, 'User state component elements should be visible'),
     {
       annotation: {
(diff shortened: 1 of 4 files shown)


tests/e2e/tests/UserStateComponent/UserStateComponentVisible.ts
```diff
@@ -10,7 +10,7 @@ export default (
   disasterType: string,
 ) => {
   // Skip for now, as test often fails
```
jannisvisser, line 12: @Piotrk39 you can remove this comment-line as well

Piotrk39: Yes good point since it's not skipped

PR #1934, opened 2025-01-22
prototype: e2e multi-disaster-type setup AB#32771

jannisvisser:
## Describe your changes

See notes in [AB#32771](https://dev.azure.com/redcrossnl/9e7ca3cc-bb97-4471-a25c-fd9787af0fe8/_workitems/edit/32771)

Initial prototype attempt to incorporate some ideas for multi-disaster-type setup for e2e
- See 'ScenarioNoTrigger.spec' only for now
- within it I loop over disaster-types
- there's an example of a test that is skipped conditional on disaster-type
- there's an example of an assertion within a test being skipped conditional on disaster-type

### DIFF
--- a/tests/e2e/tests/MapComponent/MapComponentGloFASStations.ts
+++ b/tests/e2e/tests/MapComponent/MapComponentGloFASStations.ts
@@ -3,14 +3,15 @@ import { qase } from 'playwright-qase-reporter';
 import { NoTriggerDataSet } from 'testData/testData.enum';
 
 import { Components, Pages } from '../../helpers/interfaces';
+import { DisasterTypeEnum } from '../ScenarioNoTrigger.spec';
 
 export default (
   pages: Partial<Pages>,
   components: Partial<Components>,
-  disasterType: string,
+  disasterType: DisasterTypeEnum,
 ) => {
   test(
-    qase(28, 'glofas_stations should be visible'),
+    qase(28, `glofas_stations should be visible - ${disasterType}`),
     {
       annotation: {
         type: 'url',
@@ -23,6 +24,12 @@ export default (
 
       if (!dashboard || !userState || !map) {
         throw new Error('pages and components not found');
+      }
+
+      // PROTOTYPE EXAMPLE of disaster-type specific test skip
+      if (disasterType !== DisasterTypeEnum.Floods) {
+        console.log('Skipping test for disaster type:', disasterType);
+        return;
       }
 
       // Navigate to disaster type the data was mocked for
--- a/tests/e2e/tests/MapComponent/MapComponentVisible.ts
+++ b/tests/e2e/tests/MapComponent/MapComponentVisible.ts
@@ -3,6 +3,7 @@ import { qase } from 'playwright-qase-reporter';
 import { NoTriggerDataSet } from 'testData/testData.enum';
 
 import { Components, Pages } from '../../helpers/interfaces';
+import { DisasterTypeEnum } from '../ScenarioNoTrigger.spec';
 
 export default (
   pages: Partial<Pages>,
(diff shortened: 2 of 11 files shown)


Piotrk39 (approved):
It looks good to me. I think I understand the concept. I just wonder if we still want to reuse the same 2 tests trigger and no trigger. Maybe we want to extend it with other scenarios e.g trigger-drought etc.? Or we want to loop through the reset endpoints and execute the same scenarios?

tests/e2e/tests/MapComponent/MapComponentVisible.ts
```diff
@@ -32,7 +33,11 @@ export default (
         countryName: NoTriggerDataSet.CountryName,
       });
       await map.mapComponentIsVisible();
-      await map.breadCrumbViewIsVisible({ nationalView: true });
+
+      // PROTOTYPE EXAMPLE of disaster-type specific assertion
+      if (disasterType === DisasterTypeEnum.Floods) {
```
gulfaraz, line 38: Are the breadcrumbs only used in **flood**?

We should use these features consistently for all country hazards. We can use _conditional_ tests as a crutch until we refactor these exceptions out of the portal. In the long term, such branching of logic will only get more complex thus making IBF (and its tests) unmanageable.

tests/e2e/tests/MapComponent/MapComponentGloFASStations.ts
```diff
@@ -25,6 +26,12 @@ export default (
         throw new Error('pages and components not found');
       }
 
+      // PROTOTYPE EXAMPLE of disaster-type specific test skip
```
gulfaraz, line 29: This should be conditional based on whether `glofas_stations` layer is in the scenario's `testData`. So it's not tied to a specific hazard.

My interpretation is the **visibility** of the `glofas_stations` layer is not tied to a specific hazard. _It is tied according to the data but not as the feature's behaviour_. The feature is implemented in a way the `glofas_stations` will be visible regardless of the hazard as long as the layer is enabled (in the database) and data is available. An extreme example is that the `glofas_stations` layer is enabled for any hazard including drought.

The above assumes we want the tests to test a feature and not the data flowing through the feature. It's difficult to make this distinction, especially in e2e tests.

If we stick to these conditional tests, I prefer that the conditions be checked outside the test. So the check will be done [here](https://github.com/rodekruis/IBF-system/blob/5db875b00f2451d43a7cfaad0c11330cdf33860d/tests/e2e/tests/ScenarioNoTrigger.spec.ts#L85-L86), for example
```typescript
if (disasterType === DisasterTypeEnum.Floods) {
    MapComponentGloFASStations(pages, components, disasterType); 
    MapComponentGloFASStationsWarning(pages, components, disasterType); 
}
```

tests/e2e/tests/TimelineComponent/TimelineComponentDisabled.ts
```diff
@@ -31,7 +31,7 @@ export default (
       await userState.headerComponentIsVisible({
         countryName: NoTriggerDataSet.CountryName,
       });
-      await timeline.timelineIsInactive();
+      await timeline.timelineIsInactive(); // TODO: make conditional on disaster-type
```
gulfaraz, line 34: Are the ~~breadcrumbs~~ timelines inactive only used in flood?

We should use these features consistently for all country hazards. We can use conditional tests as a crutch until we refactor these exceptions out of the portal. In the long term, such branching of logic will only get more complex thus making IBF (and its tests) unmanageable.

jannisvisser:
@gulfaraz this old PR is outdated now right? Can it be closed?

gulfaraz:
@jannisvisser yes

PR #1936, opened 2025-01-24
chore: remove enableStopTrigger AB#32660

jannisvisser:
## Describe your changes

## Notes for the reviewer


services/API-service/src/scripts/json/countries.json
```diff
@@ -440,7 +440,6 @@
         ],
         "eapLink": "https://510ibfsystem.blob.core.windows.net/about-trigger/MWI-flashfloods-about.pdf",
         "enableEarlyActions": false,
-        "enableStopTrigger": false,
```
gulfaraz, line 443: Was `enableStopTrigger` configured only for MWI flash-flood? This means all other countries and hazards were allowed to stop a forecasted trigger. Right?

services/API-service/src/scripts/json/countries.json
```diff
@@ -440,7 +440,6 @@
         ],
         "eapLink": "https://510ibfsystem.blob.core.windows.net/about-trigger/MWI-flashfloods-about.pdf",
         "enableEarlyActions": false,
```
gulfaraz, line 442: _This comment is unrelated so feel free to ignore it for now._

I'm also wondering about the implication of this `enableEarlyActions` flag.

If the list of early actions is empty for this country-hazard then this can be inferred to be false instead of this explicit config.

gulfaraz (approved):
LGTM based on the description in [this task](https://dev.azure.com/redcrossnl/IBF/_workitems/edit/32660).

services/API-service/src/scripts/json/countries.json
jannisvisser: correct

services/API-service/src/scripts/json/countries.json
jannisvisser: Valid point. I don't recall the exact reasoning for why it was put in like this at the time (and if the functionality still works now as intended then). I think we can leave this until we are touching the early actions in some way somewhere.

PR #1950, opened 2025-01-28
Update test ids and align test titles in Azure Test Plan

Piotrk39:
## Describe your changes

Updates test titles in Azure Test Plan, removes URLs to Qase and adds test ids from Test Plan

### DIFF
--- a/tests/e2e/Pages/MapComponent.ts
+++ b/tests/e2e/Pages/MapComponent.ts
@@ -129,7 +129,7 @@ class MapComponent extends DashboardPage {
   async assertAdminBoundariesVisible() {
     await this.page.waitForLoadState('networkidle');
     await this.page.waitForLoadState('domcontentloaded');
-    // await this.page.waitForTimeout(500); // This is a workaround for the issue with the map not loading in the same moment as the dom
+    await this.page.waitForTimeout(500); // This is a workaround for the issue with the map not loading in the same moment as the dom
     await this.page.waitForSelector('.leaflet-interactive');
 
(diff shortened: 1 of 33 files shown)


jannisvisser (changes requested):
@Piotrk39 not sure if this was to be reviewed, but looks mostly good. When I checkout locally and search all for 'qase' I do find some occurences still.
- one test missed it seems (MapComponentInfoPopover.ts)
- some commented out in playwright.config.ts 
- the package itself can be removed
- and I think we can also remove the Qase account we have right? (made a task)
- and lastly, I see that the URLs are removed as discussed. But this was quite handy. What is the new easiest way to get there? (Just copying the id and pasting it in DevOps searchbar works and is easy enough though I guess)

Piotrk39:
> @Piotrk39 not sure if this was to be reviewed, but looks mostly good. When I checkout locally and search all for 'qase' I do find some occurences still.
> 
> * one test missed it seems (MapComponentInfoPopover.ts)
> * some commented out in playwright.config.ts
> * the package itself can be removed
> * and I think we can also remove the Qase account we have right? (made a task)
> * I see that the URLs are removed as discussed. But this was quite handy. What is the new easiest way to get there? (Just copying the id and pasting it in DevOps searchbar works and is easy enough though I guess)
> * In Azure Test Plan I don't see which test cases have been automated already and which not right? Add as task for that as well?

@jannisvisser I will update documentation regarding the usage of Azure test plan

jannisvisser (approved):

PR #1962, opened 2025-02-05
Add test case and update trigger scenarios

Piotrk39:
## Describe your changes

Adds e2e test for validating timeline to be disabled in trigger flood scenario

### DIFF
--- a/tests/e2e/Pages/TimelineComponent.ts
+++ b/tests/e2e/Pages/TimelineComponent.ts
@@ -64,6 +64,7 @@ class TimelineComponent extends DashboardPage {
 
   async assertPurpleTimelineButtonElements() {
     await this.waitForTimelineToBeLoaded();
+    await this.page.waitForTimeout(500);
 
     const timelinePeriods = this.page.locator(
       '[data-testid="timeline-button"][ng-reflect-color="ibf-trigger-alert-secondary"]',
--- a/tests/e2e/Pages/TimelineComponent.ts
+++ b/tests/e2e/Pages/TimelineComponent.ts
@@ -70,26 +70,14 @@ class TimelineComponent extends DashboardPage {
     const timelinePeriods = this.page.locator(
       '[data-testid="timeline-button"][color="ibf-trigger-alert-secondary"]',
     );
-    // Debug logging
-    console.log('Waiting for timeline buttons to be visible...');
-    await this.page.waitForSelector('[data-testid="timeline-button"]', {
-      timeout: 5000,
-      state: 'visible',
-    });
-    console.log('Timeline buttons should be loaded.');
-
-    const html = await this.page.content();
-    console.log(html);
 
     const count = await timelinePeriods.count();
-    console.log(`Found ${count} timeline buttons.`);
-
     expect(count).toBeGreaterThan(0);
 
     for (let i = 0; i < count; i++) {
       const button = timelinePeriods.nth(i);
       const childElement = button.locator(
-        '[ng-reflect-src="/assets/icons/Alert_Title_Purp"]',
+        '[ng-reflect-src="/assets/icons/Alert_Title_Purple.svg"]',
       );
       await expect(childElement).toBeVisible();
     }
--- a/tests/e2e/Pages/TimelineComponent.ts
+++ b/tests/e2e/Pages/TimelineComponent.ts
@@ -14,9 +14,13 @@ class TimelineComponent extends DashboardPage {
     this.timeline = this.page.getByTestId('timeline-button');
   }
 
-  async timlineElementsAreVisible() {
+  async waitForTimelineToBeLoaded() {
     await this.page.waitForLoadState('domcontentloaded');
     await this.page.waitForSelector('[data-testid="timeline-button"]');
+  }
+
+  async timlineElementsAreVisible() {
+    await this.waitForTimelineToBeLoaded();
 
     const timelinePeriods = this.timeline;
     const count = await timelinePeriods.count();
@@ -28,8 +32,7 @@ class TimelineComponent extends DashboardPage {
   }
 
   async timelineIsInactive() {
-    await this.page.waitForLoadState('domcontentloaded');
-    await this.page.waitForSelector('[data-testid="timeline-button"]');
+    await this.waitForTimelineToBeLoaded();
 
     const timelinePeriods = this.timeline;
     const count = await timelinePeriods.count();
@@ -42,6 +45,8 @@ class TimelineComponent extends DashboardPage {
   }
 
   async validateTimelineDates() {
+    await this.waitForTimelineToBeLoaded();
+
     const timelinePeriods = this.timeline;
     const count = await timelinePeriods.count();
 
@@ -58,6 +63,8 @@ class TimelineComponent extends DashboardPage {
   }
 
   async assertPurpleTimelineButtonElements() {
+    await this.waitForTimelineToBeLoaded();
+
     const timelinePeriods = this.page.locator(
       '[data-testid="timeline-button"][ng-reflect-color="ibf-trigger-alert-secondary"]',
     );
--- a/tests/e2e/Pages/TimelineComponent.ts
+++ b/tests/e2e/Pages/TimelineComponent.ts
@@ -33,7 +33,7 @@ class TimelineComponent extends DashboardPage {
     }
   }
 
-  async validateTimelineBasedOnHeader({
+  async validateTimelineBasedOnDisasterName({
     disasterName,
   }: {
     disasterName: string;
--- a/tests/e2e/Pages/TimelineComponent.ts
+++ b/tests/e2e/Pages/TimelineComponent.ts
@@ -64,7 +64,7 @@ class TimelineComponent extends DashboardPage {
 
   async assertPurpleTimelineButtonElements() {
     await this.waitForTimelineToBeLoaded();
-    await this.page.waitForTimeout(500);
+    await this.page.waitForTimeout(1000);
 
     const timelinePeriods = this.page.locator(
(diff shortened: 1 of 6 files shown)


tests/e2e/tests/TimelineComponent/TimelineComponentNotClickable.ts
jannisvisser, line 1: this test is now exactly equal to the no-trigger TimelineComponentDisabled. It's weird that do exactly the same but have a different name. I get that you cannot give 2 different files a different name. Maybe just use the same test/file then in both no-trigger and trigger? Not sure what the solution is, but at least somehow change something to reflect this.  (FYI based on other comments in this PR, the situation might actually change because afterwards they're not exactly the same any more.)

tests/e2e/tests/TimelineComponent/TimelineComponentNotClickable.ts
```diff
@@ -0,0 +1,31 @@
...
+      throw new Error('pages and components not found');
+    }
+
+    // Navigate to disaster type the data was mocked for
+    await dashboard.navigateToDisasterType(disasterType);
+    // Assertions
+    await userState.headerComponentIsVisible({
+      countryName: TriggerDataSet.CountryName,
+    });
+    await timeline.timelineIsInactive();
```
jannisvisser, line 25: I would like to make a start here with thinking about this more disaster-type-agnostic. So I think you need to define a const in test-data (or new file, or wherever is good) that specifies which disaster-types have inactive timeline, and then dynamically determine here what to assert for.

tests/e2e/tests/TimelineComponent/TimelineComponentNotClickable.ts
```diff
@@ -0,0 +1,31 @@
...
+import { TriggerDataSet } from 'testData/testData.enum';
+
+import { Components, Pages } from '../../helpers/interfaces';
+
+export default (
+  pages: Partial<Pages>,
+  components: Partial<Components>,
+  disasterType: string,
+) => {
+  test('[33066] Timeline Buttons are not clickable in floods', async () => {
```
jannisvisser, line 11: Similar comment, remove the word floods here. Try to start thinking of every test in a more generic way, that it should theoretically be expandable to all disaster-types, even though not technically possible yet. Please update the Test Case also accordingly.

jannisvisser (changes requested):
I think the test case is very narrow currently and we could assert a lot more about the timeline in trigger state. E.g.
- test at least one purple button
- test at least one button with triangle icon. 
- no purples buttons without triangle icon. 
image

I think it's good if we try to be a little bit more exhaustive in things we test for, to not get a false sense of coverage. Of course, if you add the above, edit the test case / filename / etc accordingly again.

tests/e2e/tests/TimelineComponent/TimelineComponentNotClickable.ts
Piotrk39: I think it is going to be tricky to use in one test case both trigger and no trigger scenario since now we scope for two different DB resets. Therefore the best solution is to keep them separate. And indeed making changes from the PR comments the situation will change and this test will be much different.

tests/e2e/tests/TimelineComponent/TimelineComponentNotClickable.ts
Piotrk39: Changed the test title

tests/e2e/tests/TimelineComponent/TimelineComponentNotClickable.ts
Piotrk39: I have added assertions for the mentioned scenarios:

1. test at least one purple button
2. test at least one button with triangle icon.
3. no purples buttons without triangle icon.
4. date format is as expected (this is disaster-type dependent, so set this up in a dynamic way, that gets the expected date format for floods from a const)

As per making the test disaster type agnostic by selecting disaster type from the list and then assessing if it should have inactive timeline it is going to be more tricky. But I can maybe make a list and based on the header text it can assert if it should be active or inactive. Or something like that?

tests/e2e/tests/TimelineComponent/TimelineComponentNotClickable.ts
```diff
@@ -0,0 +1,35 @@
...
+import { TriggerDataSet } from 'testData/testData.enum';
+
+import { Components, Pages } from '../../helpers/interfaces';
+
+export default (
+  pages: Partial<Pages>,
+  components: Partial<Components>,
+  disasterType: string,
+) => {
+  test('[33066] Timeline Buttons are not clickable', async () => {
```
jannisvisser, line 11: with the extra assertions this test description is no longer valid. Please also update the filename + test case in Azure Test plan.

tests/e2e/testData/testData.enum.ts
```diff
@@ -20,6 +20,8 @@ export enum TriggerDataSet {
   firstName = 'Uganda',
   lastName = 'Manager',
 }
+// For now there are only floods
+export const disasterTypeWithInactiveTimeline = ['floods'];
```
jannisvisser, line 24: good step for now I think (I'm also searching still how to best go about this, let's just improve as we go along). Some small comments for now though:
- in line with the ACTIVE_LAYERS below, let's also use caps here. I believe it's standard practice for const's.
- Let's already include 'flash-floods' in the array as well, and add a note that 'drought' might be added soon.

tests/e2e/Pages/UserStateComponent.ts
```diff
@@ -44,6 +44,18 @@ class UserStateComponent extends DashboardPage {
     );
   }
 
+  async headerComponentIncludesCorrectDisasterType({
```
jannisvisser, line 47: this method is not used anywhere?

tests/e2e/Pages/TimelineComponent.ts
```diff
@@ -39,6 +57,60 @@ class TimelineComponent extends DashboardPage {
...
+    await this.waitForTimelineToBeLoaded();
+
+    const timelinePeriods = this.timeline;
+    const count = await timelinePeriods.count();
+
+    expect(count).toBeGreaterThan(0);
+
+    const today = new Date();
+    for (let i = 0; i < count; i++) {
+      const expectedDate = format(addDays(today, i), 'EEE dd MMM yyyy');
```
jannisvisser, line 90: sure, so this is floods-specific, as you commented. Fine for now, But we will run into this soon then when working on the multi-disaster-type item.

tests/e2e/Pages/TimelineComponent.ts
```diff
@@ -39,6 +57,60 @@ class TimelineComponent extends DashboardPage {
...
+    );
+
+    const count = await timelinePeriods.count();
+    expect(count).toBeGreaterThan(0);
+
+    for (let i = 0; i < count; i++) {
+      const button = timelinePeriods.nth(i);
+      const childElement = button.locator('[role="img"]');
+      await expect(childElement).toBeVisible();
+      await expect(button).not.toHaveAttribute('disabled', '');
```
jannisvisser, line 76: I realize this method is not used in practice yet, as 'floods' uses the timelineIsInactive method. But this logic here will be incorrect as it expects all timeline-buttons to not be disabled, which is not the case. Also for the other disaster-types only some relevant timeline buttons will be enabled, the rest not. Given that this logic is currently not used and incorrect, maybe just remove it and instead put a comment that explains for now that disaster-types with active timelines are not in scope for now?

tests/e2e/Pages/TimelineComponent.ts
```diff
@@ -39,6 +57,60 @@ class TimelineComponent extends DashboardPage {
...
+    const timelinePeriods = this.page.locator(
+      '[data-testid="timeline-button"][color="ibf-trigger-alert-secondary"]',
+    );
+
+    const count = await timelinePeriods.count();
+    expect(count).toBeGreaterThan(0);
+
+    for (let i = 0; i < count; i++) {
+      const button = timelinePeriods.nth(i);
+      const childElement = button.locator('[role="img"]');
```
jannisvisser, line 110: is this the line that checks if the purple button also includes a triangle icon? If so, please make it a bit more explicit, either through variable names or comments.

jannisvisser: Currently there is no assertion yet on 'test at least one button with triangle icon' right?

jannisvisser (changes requested):
Good improvements! Some questions/comments still.

tests/e2e/Pages/TimelineComponent.ts
Piotrk39: I am not sure how other disaster types are configured but we can set the same enum that it validates different date formats depending on the disaster type. We can get back to it when we start doing multi disaster type.

tests/e2e/Pages/TimelineComponent.ts
Piotrk39: It is included in a test of purple buttons since this is required element. I can change the naming so it is maybe more explicit. Or I can add a comment.

tests/e2e/tests/TimelineComponent/TimelineComponentNotClickable.ts
Piotrk39: Changed to: "Timeline's purple buttons are active or inactive dependent on disaster type"

tests/e2e/testData/testData.enum.ts
Piotrk39: Done! changed the name to caps, added flash-floods and drought in the comment

tests/e2e/Pages/TimelineComponent.ts
Piotrk39: I added a comment before the function.

jannisvisser (approved):

PR #1977, opened 2025-02-12
refactor: rename trigger-per-lead-time entity + columns AB#33395

jannisvisser:
## Describe your changes

First isolated PR that 
- renames trigger-per-lead-time table+entityclassname to alert-per-lead-time (thereby acknowledging that the data is about any alerts, not just triggers)
- renames property 'thresholdReached' to 'forecastTrigger', in line with the same change we want to make in event table
- renames property 'triggered' to 'forecastAlert', following the same naming convention, but then acknowledging this boolean indicates any alert, whereas the previous one indicates a trigger alert.
- This PR deliberatedly stays away from any further changing in dto's, variable names, etc. Let's first agree on this name before processing further. 

## Notes for the reviewer


services/API-service/src/api/event/event.service.ts
```diff
@@ -197,20 +197,20 @@ export class EventService {
...
       );
-    const triggersPerLeadTime: TriggerPerLeadTime[] = [];
+    const triggersPerLeadTime: AlertPerLeadTimeEntity[] = [];
     const timestamp = uploadTriggerPerLeadTimeDto.date || new Date();
     for (const leadTime of uploadTriggerPerLeadTimeDto.triggersPerLeadTime) {
       // Delete existing entries in case of a re-run of the pipeline within the same time period
       await this.deleteDuplicates(uploadTriggerPerLeadTimeDto);
 
-      const triggerPerLeadTime = new TriggerPerLeadTime();
+      const triggerPerLeadTime = new AlertPerLeadTimeEntity();
```
gulfaraz, line 206: > const triggerPerLeadTime = new AlertPerLeadTimeEntity();

An instance of **AlertPerLeadTimeEntity** should be called **alertPerLeadTime**.

services/API-service/src/api/event/event.service.ts
```diff
@@ -197,20 +197,20 @@ export class EventService {
         uploadTriggerPerLeadTimeDto.date,
         uploadTriggerPerLeadTimeDto.triggersPerLeadTime[0].leadTime,
       );
-    const triggersPerLeadTime: TriggerPerLeadTime[] = [];
+    const triggersPerLeadTime: AlertPerLeadTimeEntity[] = [];
```
gulfaraz, line 200: How about?
```typescript
const alertsPerLeadTime: AlertPerLeadTimeEntity[] = [];
```

services/API-service/src/api/event/event.service.ts
```diff
@@ -583,8 +583,8 @@ export class EventService {
       .select([
         'triggerPerLeadTime.leadTime as "leadTime"',
         'triggerPerLeadTime.date as date',
-        'MAX(CASE WHEN triggerPerLeadTime.triggered = TRUE THEN 1 ELSE 0 END) as triggered',
-        'MAX(CASE WHEN triggerPerLeadTime.thresholdReached = TRUE THEN 1 ELSE 0 END) as "thresholdReached"',
+        'MAX(CASE WHEN triggerPerLeadTime.forecastAlert = TRUE THEN 1 ELSE 0 END) as triggered',
```
gulfaraz, line 586: In case you missed this, the alias still uses the old name - `as triggered`

services/API-service/src/api/event/event.service.ts
```diff
@@ -583,8 +583,8 @@ export class EventService {
...
+        'MAX(CASE WHEN triggerPerLeadTime.forecastTrigger = TRUE THEN 1 ELSE 0 END) as "thresholdReached"',
```
gulfaraz, line 587: In case you missed this, the alias still uses the old name - `as thresholdReached`

services/API-service/src/api/event/trigger-per-lead-time.entity.ts
```diff
@@ -42,8 +42,8 @@ export class TriggerPerLeadTime {
   public eventName: string;
 
   @Column({ default: false })
-  public triggered: boolean;
+  public forecastAlert: boolean;
```
gulfaraz, line 45: Let's add a note to help our future selves to interpret this variable.

```
public forecastAlert: boolean; // true indicates a trigger or warning for this lead time, false indicates neither REFACTOR: to either use this for all disaster types or to remove this exception for floods
```

This looks like it's user-facing via the api so possibly update the wiki to help others also interpret this variable.

services/API-service/src/api/event/trigger-per-lead-time.entity.ts
```diff
@@ -42,8 +42,8 @@ export class TriggerPerLeadTime {
...
 
   @Column({ default: true })
-  public thresholdReached: boolean;
+  public forecastTrigger: boolean;
```
gulfaraz, line 48: Let's add a note to help our future selves to interpret this variable.

```
public forecastTrigger: boolean; // true indicates a trigger for this lead time, false indicates either a warning or no event REFACTOR: to either use this for all disaster types or to remove this exception for floods
```

This looks like it's user-facing via the api so possibly update the wiki to help others also interpret this variable.

services/API-service/src/scripts/mock.service.ts
```diff
@@ -48,8 +48,8 @@ export class Event {
 export class MockService {
   @InjectRepository(EventPlaceCodeEntity)
   private readonly eventPlaceCodeRepo: Repository<EventPlaceCodeEntity>;
-  @InjectRepository(TriggerPerLeadTime)
-  private readonly triggerPerLeadTimeRepo: Repository<TriggerPerLeadTime>;
+  @InjectRepository(AlertPerLeadTimeEntity)
+  private readonly triggerPerLeadTimeRepo: Repository<AlertPerLeadTimeEntity>;
```
gulfaraz, line 52: How about?

`private readonly alertPerLeadTimeRepo: Repository<AlertPerLeadTimeEntity>;`

services/API-service/src/shared/helper.service.ts
```diff
@@ -94,8 +94,9 @@ export class HelperService {
     countryCodeISO3: string,
     disasterType: DisasterType,
   ): Promise<DateDto> {
-    const triggerPerLeadTimeRepository =
-      this.dataSource.getRepository(TriggerPerLeadTime);
+    const triggerPerLeadTimeRepository = this.dataSource.getRepository(
```
gulfaraz, line 97: `const alertPerLeadTimeRepository = this.dataSource.getRepository(AlertPerLeadTimeEntity);`

gulfaraz (changes requested):
**Alert** doesn't match the existing definition in the glossary. The glossary uses alert as a _verb_ while this PR uses it as a _noun_.

**Risk** or **exposure** feels more appropriate here. As warning and trigger describe the exposure or risk of a place/time.

services/API-service/src/api/event/event.service.ts
jannisvisser: very true, but this is what I meant when saying let's first agree on the general rename before I process all of this.

services/API-service/src/api/event/event.service.ts
jannisvisser: same

services/API-service/src/api/event/event.service.ts
jannisvisser: same

services/API-service/src/api/event/event.service.ts
jannisvisser: same

services/API-service/src/scripts/mock.service.ts
jannisvisser: same

services/API-service/src/shared/helper.service.ts
jannisvisser: same

services/API-service/src/api/event/trigger-per-lead-time.entity.ts
jannisvisser: commend included
this particular entity property itself is not yet user-facing right? There's also no existing wiki documentation on this atm. I will add a link to this PR comment in the documentation task of this item.

services/API-service/src/api/event/event.service.ts
jannisvisser: after seeing all other comments, I think only point to agree on here is the general use of the term 'alert' right?
- I do indeed interpret alert as a noun and as the general term that covers warning (noun) and trigger (noun). E.g. in en.json, you constantly see it used as a noun.
- I do not assume the glossary to be up-to-date. I don't recognize all of the current content in it and don't consider it to be team-agreed-upon. I have marked it as part of the standard documentation to start updating, and this is a good item to do an iteration on that. I will make a specific task for it.
- about the alternatives: 'exposure' I find too tricky, as it mixes up with the 'mainExposureIndicator'/population_affected, which has different uses again. 'risk' is a term I personally have never used or heard used much in this context.
- So I do actually propose to start using 'alert' (noun) as the word that covers warnings and triggers. Only yesterday in the call with Gal about emails I heard this confirmed from her side as well.

gulfaraz (approved):

PR #1982, opened 2025-02-17
Feat.process pipeline run old and new

jannisvisser:
## Describe your changes

- Builds upon #1981 
- See details in [AB#33398](https://dev.azure.com/redcrossnl/9e7ca3cc-bb97-4471-a25c-fd9787af0fe8/_workitems/edit/33398)
- PR has failing tests, these are fixed in follow-up PR #1989 
- This PR has merge conflicts. Before rebasing first review and merge PR #1988 and #1989 into here.


services/API-service/src/scripts/mock.service.ts
```diff
@@ -189,10 +190,24 @@ export class MockService {
         }
 
         if (this.shouldMockAlertPerLeadTime(disasterType)) {
+          if (MOCK_USE_OLD_PIPELINE_UPLOAD) {
```
gulfaraz, line 193: @jannisvisser should this IF have an ELSE block?

gulfaraz (approved):

services/API-service/src/scripts/mock.service.ts
jannisvisser: @gulfaraz yes, the part below it should be in else. I guess I wanted to go for an 'early return' type of coding here, but that doesn't work if you're not actually returning/continuing from something to skip the code below it.. Can you change it?

gulfaraz:
@jannisvisser In the new setup, the mock service requires `alert-per-lead-time` table to be populated for non-flood disaster types. I believe the integration tests will pass if these files are added and mock service is updated to populate `alert-per-lead-time` table for all disaster types.

jannisvisser:
> @jannisvisser In the new setup, the mock service requires `alert-per-lead-time` table to be populated for non-flood disaster types. I believe the integration tests will pass if these files are added and mock service is updated to populate `alert-per-lead-time` table for all disaster types.

I'm not getting this. 
- The integration tests were all running fine in earlier iterations of this work, when did they start to fail?
- Also, alert-per-lead-time should be filled (and was filled) just through the /events/process endpoint, which contains the insertAlertPerLeadTime, and this doesn't differ between disaster-types. The separate endpoint+mock-method for filling alert-per-lead-time is just for the additional (non-event-starting) lead-times, and that should be indeed only for floods.
- I'm sure there just some small mistake going on again, like last time with the e2e tests. I can help with figuring out what if needed. Because it's hard to tell like this.

services/API-service/src/api/event/event.service.ts
```diff
@@ -481,11 +484,11 @@ export class EventService {
...
         'disaster."mainExposureIndicator" as "exposureIndicator"',
         'event."mainExposureValue" as "exposureValue"',
-        `CASE 
-        WHEN event."forecastTrigger" = true THEN 'Trigger' 
-        WHEN event."forecastSeverity" = 1 THEN 'High warning' 
-        WHEN event."forecastSeverity" = 0.7 THEN 'Medium warning' 
-        WHEN event."forecastSeverity" = 0.3 THEN 'Low warning' 
+        `CASE
+        WHEN event."forecastTrigger" = true THEN 'Trigger'
+        WHEN event."forecastSeverity" = 1 THEN 'High warning'
```
jannisvisser, line 494: Hmm, this is my change, but realizing that this code was and is suboptimal, but mostly in current practice this  will go wrong because drought/flash-flood/typhoon warnings will all have forecastSeverity=1, while for floods the 'high warning' will for now not be implented (or if so, then include this in AB#33695). So for now changing this to 'Warning' to at least better reflect the current situation with this suboptimal code.

PR #2013, opened 2025-02-26
test: refactors to make tests less flaky

gulfaraz:
## Describe your changes

I created this PR only after all the tests ran 3 consecutive times without any failures. It includes,
1. Use specific selectors to identify UI elements. For example, use `.admin-boundary` instead of `.leaflet-interactive` to hover over admin boundary polygons. It used to include GloFAS markers as they are also interactive leaflet elements.
2. Modify tests so as not to skip necessary UI actions. In the case of opening info popups, the test should explicitly close the popup after assertions by clicking on the backdrop.
3. Remove most timeouts. The timeouts did not reduce the flakiness. A few timeouts remain because I had to stop refactoring tests at some point.

I'm now certain that flaky tests can be completely avoided. The reasons for flakiness are varied. My understanding of the user interface significantly helped identify some causes. Playwright also has unexpected behaviours based on how and where we use locators. Some tests became stable by defining the locator at the time of the assertion (instead of the page constructor). We're learning this together and it looks promising that we will achieve a suite of stable test cases which scale across countries and hazards.

@Piotrk39 For future tests, please check with me before you add any timeouts or selectors/locators. I noticed code evidence of the impossible task we're putting you through. Your work so far is impressive, well done 👏 🏆 🥇

### DIFF
--- a/interfaces/IBF-dashboard/src/app/components/aggregates/aggregates.component.html
+++ b/interfaces/IBF-dashboard/src/app/components/aggregates/aggregates.component.html
@@ -3,7 +3,6 @@
     <ion-row
       class="component-title"
       style="background: var(--ion-color-ibf-secondary)"
-      data-testid="action-title"
     >
       <ion-col class="ion-no-padding" size-lg="11" size-md="11" size-xs="11">
         <ion-item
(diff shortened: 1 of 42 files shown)


jannisvisser:
@gulfaraz amazing!

Piotrk39:
Great work @gulfaraz !!!

PR #2028, opened 2025-03-04
fix: change close-events api path typo AB#34185

jannisvisser:
## Describe your changes

Floods pipeline failed because api path changed from /event/close-events to /events/close-events (plural in first part).
This is changed in this PR.

### DIFF
--- a/services/API-service/src/api/process-pipeline/process-pipeline.controller.ts
+++ b/services/API-service/src/api/process-pipeline/process-pipeline.controller.ts
@@ -43,7 +43,7 @@ export class ProcessPipelineController {
     status: 201,
     description: 'Closed finished events.',
   })
-  @Post('events/close-events')
+  @Post('event/close-events')
   public async closeEvents(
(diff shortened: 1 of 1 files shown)


services/API-service/src/api/process-pipeline/process-pipeline.controller.ts
```diff
@@ -43,7 +43,7 @@ export class ProcessPipelineController {
     status: 201,
     description: 'Closed finished events.',
   })
-  @Post('events/close-events')
+  @Post('event/close-events')
```
gulfaraz, line 46: @jannisvisser I thought this was intentional.

jannisvisser: Well it was somewhat, in the sense of changing all paths in process-pipeline.controller to /events. I just didn't realize here that this endpoint was kept in for the transition period specifically and thus needed an unchanged api path.

PR #2030, opened 2025-03-05
feat: popups to set trigger AB#34114

jannisvisser:
## Describe your changes

- Implements set trigger popover + confirmation popover.
- See [AB#34114](https://dev.azure.com/redcrossnl/9e7ca3cc-bb97-4471-a25c-fd9787af0fe8/_workitems/edit/34114)
- Front-end does not look optimal / like design yet. Leaving that for later Design Review task.


gulfaraz (changes requested):
I used [this design](https://www.figma.com/design/V0TaRkCWMeqULWka1yPn33/%F0%9F%8F%9C%EF%B8%8F-%F0%9F%87%B1%F0%9F%87%B8-IBF-Lesotho-drought?node-id=6686-18884&t=wlrLrEQoFzrxjsae-4) as a reference for my review,

Dashboard | Design | Review
---|---|---
Screenshot 2025-03-10 at 13 07 56 | Screenshot 2025-03-10 at 13 00 57 | <ol> <li> _Set trigger_ button is not clearly visible: because of the background color which matches the parent background. </li> <li> The button is missing a border color. </li> <li> On large screens, the button is too wide: it needs a `max-width` or similar limit. </li> <li> Label is missing prefix - _Ongoing warning:_ </li> <li> Border is different but not relevant for this PR. </li> <li> Probability is missing: I believe we aligned on this with design already. </li> </ol>
Screenshot 2025-03-10 at 13 01 39 | Screenshot 2025-03-10 at 13 01 59 | <ol> <li> The dimensions of the popup are too small. This will not work when the list of warned districts increases. </li> <li> The spacing between districts and checkbox is excessive: did you check what it looks like when there are more than 10 warned districts? </li> <li> Exposed population is not shown in the popup. </li> <li> _3 districts selected_ is missing. </li> <li> The copy is different: I will assume this is fine. </li> <li> The color doesn't match the design: I assume this is also fine. </li> </ol>
Screenshot 2025-03-10 at 13 02 39 | Screenshot 2025-03-10 at 13 02 19 | <ol> <li> The dimensions of the popup are too small for the content. </li> <li> Content is aligned in columns instead of rows.</li> <li> The disclaimers are not in a bullet list. </li> <li> Checkbox instead of password: I choose to assume this is fine. </li> <li> 3 districts selected is missing. </li> <li> The copy is different: I assume this is fine. </li> <li> The colour is different: I assume this is fine. </li> </ol>

It is understandable to see differences between the design input and resulting UI, but the differences here are significant.

The most critical review applies to both the dashboard and the design. We should not trigger a second popup from within a popup. In some cases, nested popups are unavoidable. In this case, we can avoid this complex design by combining the two popups. [User experience on Stack Exchange](https://ux.stackexchange.com/a/52074/69656). I missed this during the design review, I don't remember if we had one. Maybe I overlooked this as there was confusion if we will even do this project.

interfaces/IBF-dashboard/src/app/components/set-trigger-confirm-popover/set-trigger-confirm-popover.component.scss
```diff
@@ -0,0 +1,3 @@
+ion-label {
```
gulfaraz, line 1: Use tailwind's [text-xs](https://tailwindcss.com/docs/font-size) instead.

interfaces/IBF-dashboard/src/app/components/set-trigger-popover/set-trigger-popover.component.scss
```diff
@@ -0,0 +1,3 @@
+ion-label {
```
gulfaraz, line 1: Use tailwind's [text-xs](https://tailwindcss.com/docs/font-size) instead.

interfaces/IBF-dashboard/src/global.scss
```diff
@@ -318,3 +318,8 @@ app-tooltip {
 .font-size-12 {
   font-size: 12px;
 }
+
+a {
```
gulfaraz, line 322: Use tailwind's [underline](https://tailwindcss.com/docs/text-decoration-line#underling-text) and [colour](https://tailwindcss.com/docs/colors#customizing-your-colors) instead.

jannisvisser:
@gulfaraz fully right on here and there substantial differences with design. As I said in the PR description, I was opting to leave details like this for a separate task. Can we maybe separate what is must-fix within this PR, and what can be left (for later task) or completely? (Best in a call I guess)

jannisvisser:
must-fix in this PR
- chat: 1, 2 (3 if easy)
- popovers: combine 2 popovers in 1
- popover 1: 1,2 (test with 10 districts (or even more) to test responsiveness to volume) and 3,4
- popover 2: 1,2,3,5
- tailwind comments

interfaces/IBF-dashboard/src/app/components/set-trigger-confirm-popover/set-trigger-confirm-popover.component.scss
jannisvisser: closing this, as this popover will go

interfaces/IBF-dashboard/src/global.scss
jannisvisser: does this require me to add class="underline" (and something about colour) separately to all <a> elements in the codebase? That seems unhandy. Am I missing something?

jannisvisser:
@gulfaraz I processed all the discussed points now
- only the spacing between area-checkboxes is not done as already discussed. and the scrollability there not optimal.
- and I did not process  the open tailwind comment yet with the question to you

PR #2031, opened 2025-03-05
feat: rename stopped to userTrigger

gulfaraz:
## Describe your changes

[AB#34109](https://dev.azure.com/redcrossnl/IBF/_workitems/edit/34109)

This PR includes,
1. Rename `stopped` to `userTrigger`
2. Rename `manuallyStoppedDate` to `userTriggerDate`
3. New API `set-trigger`
4. Remove old API `toggle-stopped-trigger`

Tests and docs will be updated in [AB#34121](https://dev.azure.com/redcrossnl/IBF/_workitems/edit/34121) and [AB#34123](https://dev.azure.com/redcrossnl/IBF/_workitems/edit/34123) respectively.

### DIFF
--- a/services/API-service/src/api/event/event.controller.ts
+++ b/services/API-service/src/api/event/event.controller.ts
@@ -165,10 +165,10 @@ export class EventController {
 
   @UseGuards(RolesGuard)
   @Roles(UserRole.DisasterManager)
-  @ApiOperation({ summary: 'Set trigger for event areas.' })
+  @ApiOperation({ summary: 'Set trigger for event admin-areas.' })
   @ApiResponse({
     status: 201,
-    description: 'Event areas are set to trigger.',
+    description: 'Event admin-areas are set to trigger.',
   })
   @Post('set-trigger')
   public async setTrigger(
--- a/services/API-service/src/api/event/event.controller.ts
+++ b/services/API-service/src/api/event/event.controller.ts
@@ -24,7 +24,7 @@ import { UserRole } from '../user/user-role.enum';
 import { UserDecorator } from '../user/user.decorator';
 import {
   ActivationLogDto,
-  EventPlaceCodeDto,
+  EventPlaceCodesDto,
 } from './dto/event-place-code.dto';
 import { LastUploadDateDto } from './dto/last-upload-date.dto';
 import {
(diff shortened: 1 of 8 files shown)


services/API-service/src/api/event/event.service.ts
```diff
@@ -1075,8 +1070,6 @@ export class EventService {
         eventArea.mainExposureValue = +activeAlertArea.mainExposureValue;
         eventArea.firstIssuedDate = lastUploadDate.timestamp;
         eventArea.endDate = lastUploadDate.timestamp;
-        eventArea.stopped = false;
-        eventArea.manualStoppedDate = null;
```
jannisvisser, line 1079: I guess userTrigger and date are simply filled with default values?

services/API-service/src/api/event/event.controller.ts
```diff
@@ -165,24 +165,17 @@ export class EventController {
 
   @UseGuards(RolesGuard)
   @Roles(UserRole.DisasterManager)
-  @ApiOperation({ summary: 'Stop trigger for given admin-area.' })
+  @ApiOperation({ summary: 'Set trigger for event areas.' })
```
jannisvisser, line 168: event areas is a bit of a complicated term, since we have also reserved that for the one-per-event-polygons as defined in event-area.service (see also glossary). I know that we use 'event area' already as you mean it here, also in other places in event.service but we should clear this up from now on if possible.

services/API-service/src/api/event/event.service.ts
```diff
@@ -493,11 +495,14 @@ export class EventService {
         'disaster."mainExposureIndicator" as "exposureIndicator"',
         'event."mainExposureValue" as "exposureValue"',
         `CASE
+        WHEN event."userTrigger" = true THEN 'Trigger'
```
jannisvisser, line 498: fine for now, although this will quickly change again with the calculation of 'trigger' based on 'userTrigger' and 'forecastTrigger'

jannisvisser (changes requested):
quick initial code review. I have not tested functionally. I will review more on friday.

services/API-service/src/api/event/event.service.ts
gulfaraz: @jannisvisser yes

services/API-service/src/api/event/event.service.ts
```diff
@@ -615,28 +620,18 @@ export class EventService {
...
-      const errors = 'User not found';
-      throw new HttpException({ errors }, HttpStatus.NOT_FOUND);
-    }
-    const eventPlaceCode = await this.eventPlaceCodeRepo.findOne({
-      where: { eventPlaceCodeId: eventPlaceCodeDto.eventPlaceCodeId },
-    });
-    if (!eventPlaceCode) {
-      const errors = 'Event placeCode not found';
-      throw new HttpException({ errors }, HttpStatus.NOT_FOUND);
-    }
```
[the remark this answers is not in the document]
gulfaraz: Depends on how we want to handle this, my assumptions,
1. This API is only used by the dashboard, which is under our control.
2. The API expects specific ids so malicious attack risk is minimal because the UUIDs are obscure.

In the case of invalid ids, no rows will be updated. Or the update method will throw an error, I have not tested this.
If we handle this explicitly, we have to make more decisions like how to handle partially correct requests where some ids are valid and others not. I expect typeorm will only update matching rows and ignore the invalid ids. Or throw an error so either way they work in our favour. Also, I have not tested this.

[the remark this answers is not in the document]
jannisvisser: Ok, it sounds fine. Functional testing of different scenarios should be done before we can merge this though.

services/API-service/src/api/event/event.controller.ts
jannisvisser: keeping this open, so I can check the change later

gulfaraz: @jannisvisser should I change the string back to admin area? or how do we clear this up?

jannisvisser: I'm also not sure, but maybe 'event admin-area' is at least better.

jannisvisser: I committed this

jannisvisser (approved):

PR #2033, opened 2025-03-05
test: add drought no trigger and fix failing tests AB#34224

jannisvisser:
## Describe your changes

This PR
- adds drought no trigger scenario
- fixed failing tests after doing so, by 
  - making the assertions disaster-type agnostic via test data (mostly in AggregatesComponent)
  - skipping tests based on disaster-type
  - skipping assertion based on disaster-type
- added drought warning scenario now as well (NOTE: I think that's more interesting to add than the trigger scenario. And adding both is overkill for now)
  - fixed 2 failing tests here by making them generic  
  - removed 1 flaky but irrelevant assertion

### DIFF
--- a/interfaces/IBF-dashboard/src/app/components/chat/chat.component.html
+++ b/interfaces/IBF-dashboard/src/app/components/chat/chat.component.html
@@ -107,7 +107,7 @@
               "
             ></ion-label
             ><br />
-            @if (eventState?.event?.forecastTrigger) {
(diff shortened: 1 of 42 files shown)


jannisvisser:
@gulfaraz this is ready as far as this item goes. My main open question is if the way we conditionally run test on disasterType and/or scenario doesn't get too messy already.

Current commit has 2 out of 94 tests failing due to flakiness.

PR #2036, opened 2025-03-07
feat: trigger

gulfaraz:
## Describe your changes

[AB#34110](https://dev.azure.com/redcrossnl/IBF/_workitems/edit/34110)

This PR includes,
1. Calculate **Alert Level** based on forecast severity, forecast trigger, and user trigger
2. Replace forecastTrigger with alert level in events, admin areas, and notifications.

## Notes for the reviewer

`AlertsPerLeadTime` uses `forecastTrigger` but I didn't touch that as I couldn't decipher how it should work. It appeared sufficiently isolated to keep out of this PR.

### DIFF
--- a/services/API-service/src/api/event/event.service.ts
+++ b/services/API-service/src/api/event/event.service.ts
@@ -1100,7 +1100,6 @@ export class EventService {
       endDate: LessThan(lastUploadDate.timestamp), // If the area was not prolonged earlier, then the endDate is not updated and is therefore less than the lastUploadDate
       adminArea: { countryCodeISO3 },
       disasterType,
-      userTrigger: false, // do not close event triggered by user
       closed: false,
     };
     const expiredEventAreas = await this.eventPlaceCodeRepo.find({ where });


services/API-service/src/api/event/dto/event-place-code.dto.ts
```diff
@@ -50,6 +51,9 @@ export class ActivationLogDto {
   @ApiProperty({ example: 'Maximum alert' })
   public alertClass: string;
 
+  @ApiProperty({ default: AlertLevel.NONE })
+  public alertLevel: AlertLevel;
```
jannisvisser, line 55: is this overlapping with alertClass? Is there opportunity to drop alertClass now?

services/API-service/src/api/event/event.service.ts
```diff
@@ -162,15 +161,27 @@ export class EventService {
...
+      return AlertLevel.TRIGGER;
+    }
+
+    if (event.forecastSeverity > 0.7) {
+      return AlertLevel.WARNING;
+    } else if (event.forecastSeverity > 0.3) {
+      return AlertLevel.WARNINGMEDIUM;
+    } else if (event.forecastSeverity > 0) {
+      return AlertLevel.WARNINGLOW;
+    }
```
jannisvisser, line 180: This works for now, but will need updating as we go along (which is fine). I'm not particularly happy with this definition of WARNING being placed "above" low/medium warning. But again, it works for now, so good enough. We will already touch upon this when we would facilitate high-warning in floods.

services/API-service/src/api/event/event.service.ts
```diff
@@ -501,6 +516,14 @@ export class EventService {
         WHEN event."forecastSeverity" = 0.7 THEN 'Medium warning'
         WHEN event."forecastSeverity" = 0.3 THEN 'Low warning'
         END as "alertClass"`,
+        `CASE
```
jannisvisser, line 519: same question. we can probably lose alertClass here right?

services/API-service/src/api/event/event.service.ts
```diff
@@ -1086,19 +1107,20 @@ export class EventService {
       endDate: LessThan(lastUploadDate.timestamp), // If the area was not prolonged earlier, then the endDate is not updated and is therefore less than the lastUploadDate
       adminArea: { countryCodeISO3 },
       disasterType,
+      userTrigger: false, // do not close event triggered by user
```
jannisvisser, line 1110: Hmm, I know where this comes from. But it needs to close at some point right? This way it can never disappear.

services/API-service/src/api/event/event.service.ts
```diff
@@ -1109,29 +1131,35 @@ export class EventService {
...
-      if (eventForecastTrigger) {
+      if (alertLevel === AlertLevel.TRIGGER) {
         return {
           key: EapAlertClassKeyEnum.max,
           label: 'Trigger',
           color: 'ibf-glofas-trigger',
           value: 1,
         };
-      } else if (eventForecastSeverity > 0) {
+      } else if (ALERT_LEVEL_WARNINGS.includes(alertLevel)) {
```
jannisvisser, line 1162: I think this 'if' is particulary reserved for the general 'warning' class, and not for the low/medium warning classes? As I think in those cases they would need to be specified in country-config (like currently for floods).

jannisvisser (changes requested):
I did an initial quick review with some comments. It touches quite some core code, so will look at it more in-depth later. But feel free to already look at these comments now.

services/API-service/src/api/event/event.controller.ts
```diff
@@ -174,8 +176,8 @@ export class EventController {
   public async setTrigger(
     @UserDecorator('userId') userId: string,
     @Body() eventPlaceCodesDto: EventPlaceCodesDto,
-  ): Promise<void> {
-    await this.eventService.setTrigger(userId, eventPlaceCodesDto);
+  ): Promise<UpdateResult> {
```
jannisvisser, line 179: I guess this made sense, after functionally testing the just merged PR on ibf-test. Agree?

services/API-service/src/api/event/event.service.ts
jannisvisser: @gulfaraz thought about this more, and I think we should revert this for now. To bring it back in AB#32053, where we can apply filtering such that event with userTrigger:true does close if lastUploadDate>endDate, and otherwise not. Agree?

services/API-service/src/api/event/dto/event-place-code.dto.ts
gulfaraz: > is this overlapping with alertClass?

I created a mapping from alert level to alert class [here](https://github.com/rodekruis/IBF-system/blob/6c490c442aef00ae742f315ca89f112e9410a22c/services/API-service/src/api/event/event.service.ts#L1137-L1143) to avoid the overlap. `alertClass` seems to be used in many ways so I didn't want to remove it in this PR.

> Is there opportunity to drop alertClass now?

Yes, if we refactor the _many ways_ we use alertClass.

services/API-service/src/api/event/event.service.ts
gulfaraz: [same answer](https://github.com/rodekruis/IBF-system/pull/2036#discussion_r1986956365)

services/API-service/src/api/event/event.service.ts
gulfaraz: I don't like it either but I followed the definition from [here](https://github.com/rodekruis/IBF-system/blob/fa0266bc9d645d0c35cf15cd75bcfe16233884dd/services/API-service/src/api/event/event.service.ts#L495-L500).

I prefer TRIGGER > HIGH WARNING > WARNING > LOW WARNING > NONE

services/API-service/src/api/event/dto/event-place-code.dto.ts
jannisvisser: Losing it in general may be more difficult, but this use case in particular is about including this in the activation-log, which is just one particular use that we can get rid of right? Given that in practice the values end up the same?

services/API-service/src/api/event/event.service.ts
gulfaraz: The code in this block uses code related to medium warnings so I'm not sure what you mean.

https://github.com/rodekruis/IBF-system/blob/6c490c442aef00ae742f315ca89f112e9410a22c/services/API-service/src/api/event/event.service.ts#L1163-L1168

services/API-service/src/api/event/event.service.ts
jannisvisser: let's leave it for now, and just touch upon it again in the High Warning item (or whenever it comes up more pressing).

services/API-service/src/api/event/event.controller.ts
gulfaraz: I agree

services/API-service/src/api/event/event.service.ts
jannisvisser: closing this convo in favour of the other one.

services/API-service/src/api/event/dto/event-place-code.dto.ts
gulfaraz: Yes, we can get rid of it in the activation log. I'll remove it.

gulfaraz: done https://github.com/rodekruis/IBF-system/pull/2036/commits/ff1cd23ffcb44249d268fe4d8dccebfdee515934

services/API-service/src/api/event/event.service.ts
gulfaraz: done https://github.com/rodekruis/IBF-system/pull/2036/commits/421222f1e75559ff6f4b248794ba072b0afc158e

services/API-service/src/api/event/event.service.ts
jannisvisser: Yeah, I guess you're right. And either way, it will functionally work for now. It's just another reminder of how ugly this code setup around EapAlertClass is now. We already have AB#34026 for that.. but that doesn't mean much practically unless we can tie it to some feature/project item..

gulfaraz: @jannisvisser can we tie it to this feature/project and fix it after the core work is complete?

jannisvisser: I would be happy to if possible, but I guess you understand that this project's budget is likely largely depleted after the core work. Best we can do maybe is to tie it to QA work and do combined refactoring/docs/testing on this with our AA-day testing budget. I will at least drag it up a bunch for prioritization and add the refine tag.

jannisvisser:
Additional functional testing
- mocked warning LSO, then set trigger for both areas via Swagger. Portal correctly looks triggered (map + chat, see screenshot)
- timeline and overall styling of dashboard do not look triggered yet, but this is expected as that is currently based on forecastTrigger and not on "trigger" for which we have separate task AB#34111
- I checked other disaster types for both warning and trigger events and all looks good. 
- All in all, good enough for now. We will touch upon more of the intricate working of trigger/forecastTrigger/userTrigger in follow-up tasks within this item anyway. But this looks good to merge.
image

jannisvisser (approved):

PR #2059, opened 2025-03-17
refactor: calculate event aggregates in front-end AB#34119

jannisvisser:
## Describe your changes

- Rough idea for moving the event-aggregates calculation to the front-end as discussed. 
- By making GET /alert-areas content a part of GET /event.
- This is not fully ready, as - if agreed - it needs some more cleaning up. Especially in the alert-area.service, as there is not really a reason to add a subscription to AlertAreas any more, if this data is already available earlier in eventState.
- to do also: fix integration test
- tied this to separate task [AB#34525](https://dev.azure.com/redcrossnl/9e7ca3cc-bb97-4471-a25c-fd9787af0fe8/_workitems/edit/34525)

### DIFF
--- a/interfaces/IBF-dashboard/src/app/components/event-speech-bubble/event-speech-bubble.component.ts
+++ b/interfaces/IBF-dashboard/src/app/components/event-speech-bubble/event-speech-bubble.component.ts
@@ -76,11 +76,29 @@ export class EventSpeechBubbleComponent implements AfterViewChecked, OnDestroy {
 
     if (this.event) {
       this.event.header = this.getHeader(this.event);
+      this.event.mainExposureValueSum = this.getEventMainExposureValue(
+        this.event,
+        this.mainExposureIndicatorNumberFormat,
+      );
     }
   }
 
   ngOnDestroy() {
     this.placeCodeHoverSubscription.unsubscribe();
+  }
+
+  private getEventMainExposureValue(
+    event: EventSummary,
+    mainExposureIndicatorNumberFormat: NumberFormat,
+  ) {
+    const sum = event.alertAreas.reduce(
+      (acc, alertArea) => acc + alertArea.mainExposureValue,
+      0,
+    );
+    if (mainExposureIndicatorNumberFormat === NumberFormat.perc) {
+      return sum / event.alertAreas.length;
+    }
+    return sum;
   }
 
   public selectArea(area: AlertArea) {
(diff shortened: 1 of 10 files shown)


jannisvisser:
@gulfaraz FYI. This is roughly the idea I had in mind. I think it works well in terms of losing the complicated getNrAlertAreasMainExposureValueSum method, and in exchange not really needing to add much or complex code. As a side result it completely gets rid of a full endpoint. 

The work to do here is mostly (as touched upon above) to clean up the subscription chain around this dropped endpoint.

interfaces/IBF-dashboard/src/app/components/event-speech-bubble/event-speech-bubble.component.ts
```diff
@@ -76,13 +76,31 @@ export class EventSpeechBubbleComponent implements AfterViewChecked, OnDestroy {
...
 
+  private getEventMainExposureValue(
+    event: EventSummary,
+    mainExposureIndicatorNumberFormat: NumberFormat,
+  ) {
+    const sum = event.alertAreas.reduce(
+      (acc, alertArea) => acc + alertArea.mainExposureValue,
+      0,
+    );
+    if (mainExposureIndicatorNumberFormat === NumberFormat.perc) {
```
gulfaraz, line 98: This is still a temporary solution so it needs the comment that we should not be averaging percentages.

gulfaraz (commented):
@jannisvisser Nice! The event areas on the frontend seems to work. Should the aggregates component show `1 exposed district` and `Exposed population: 500`?

LSOdroughtaggregates

interfaces/IBF-dashboard/src/app/components/event-speech-bubble/event-speech-bubble.component.ts
jannisvisser: correct, I will add it.

jannisvisser:
@gulfaraz given the work done on this already, do you think it's feasible to complete this in the current item? It would be a shame to stash this again to (probably not) come back to later..

gal-agmon:
> > do you think it's feasible to complete this in the current item?
> 
> Unfortunately, no.
> 
> Check with @gal-agmon if this is enough.

@gulfaraz I an mot sure what is "this" referring to

gulfaraz:
@gal-agmon We'll let you know once @jannisvisser and I figure out what yes is.

interfaces/IBF-dashboard/src/app/components/event-speech-bubble/event-speech-bubble.component.ts
jannisvisser: done

PR #2077, opened 2025-03-21
fix: links should open in new tab and show unbroken dead state

gulfaraz:
## Describe your changes

This PR includes,
- External links should open in a new tab, requested in design review [AB#34124](https://dev.azure.com/redcrossnl/IBF/_workitems/edit/34124) [AB#34673](https://dev.azure.com/redcrossnl/IBF/_workitems/edit/34673) https://github.com/rodekruis/IBF-system/commit/f0eda392d2b4d85dcc3dc1eb6b297d4083f33844
- Dashboard should not show a broken state with a degraded API-service.
- CI workflow should run when a PR is ready for review.

## Notes for the reviewer

Since 2021, browsers implicitly set `rel='noopener'` for anchor tags with `target='_blank'`. So I removed them across the dashboard.

> Note: Setting target="_blank" on \<a\>, \<area\> and \<form\> elements implicitly provides the same rel behavior as setting rel="noopener" which does not set window.opener.

[Source](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noopener)

In this PR, I waited for the checks to be completed. Then I marked the PR as ready for review. I expected to see that the e2e and integration tests would trigger but they didn't. So I updated the workflow.yml file. We should check if the behaviour changes as a result in future PRs.


.github/workflows/workflow-pull_request.yml
```diff
@@ -4,6 +4,11 @@ on:
   pull_request:
     branches:
       - master
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
```
jannisvisser, line 11: is this supposed to filter out 'draft'?
- is this list complete then, and not missing any?
- isn't this duplicate with the check on draft down below?

jannisvisser (approved):
see 1 question, for the rest approved.

as discussed, please:
- remove the excel and related content
- create item for your idea (to later add to the epic)

.github/workflows/workflow-pull_request.yml
gulfaraz: @jannisvisser This change is supposed to run the workflow when the _status changes_ from draft to ready for review.
The other check is to skip the job if the _status is_ draft.

We have to check if this works as intended which is only possible across PRs on GitHub.

I got this solution from https://github.com/reviewdog/action-eslint/issues/29#issuecomment-985939887

PR #2078, opened 2025-03-21
fix: only run tests on PRs to master

jannisvisser:
## Describe your changes

## Notes for the reviewer


jannisvisser:
@gulfaraz FYI. When we talked earlier I was under the impression that your change on draft PRs was about 'not reporting on the test' instead of 'not running the test'. I wanted to do the same for sub-feature-branch-PRs. But when I saw that your change actually meant not running the test, I decided to do the same for sub-feature-branch-PRs, as I think that's fine anyway.

PR #2085, opened 2025-03-24
Update dataSets to accommodate default layers for: Malaria, Typhoon and Drought

Piotrk39:
## Describe your changes

Instead of adding 3 more test cases for testing default layers in Malaria, Typhoon and Drought the below was done:
1. Uganda drought dataset was updated (UGA)
2. Malaria Trigger dataset was added (ETH)
3. Typhoon Trigger dataset was added (PHL)

## Notes for the reviewer

By adding new datasets some test cases started falling as functions were not ready to process different IBF states from Malaria and Typhoon. Here is the list of failing tests:

Dataset 75: [ethiopia@redcross.nl] ETH malaria trigger

DashboardPage

[33012] Dashboard page elements should be visible
MapComponent

[33013] Map component elements should be visible
AggregatesComponent

[33058] Aggregates component elements should be visible
[33060] Info button(s) should be clickable
ChatComponent

[33717] Click on area in triggered areas list in chat
[33056] Number of events should match the number of events in aggregates component
Note: As seen in the code, this test is failing because "IN ETHIOPIA MALARIA TRIGGER ONE BUTTON IS DISABLED BECAUSE THE TRIGGERED ARE IS ALREADY PRESENT AND THAT IS WHY THE COUNT IS 1 LESS"

Dataset 76: [philippines@redcross.nl] PHL typhoon trigger:

DashboardPage

[33012] Dashboard page elements should be visible
MapComponent

[33013] Map component elements should be visible
[33014] Map component should be interactive
AggregatesComponent

[33058] Aggregates component elements should be visible
[33060] Info button(s) should be clickable
ChatComponent

[33717] Click on area in triggered areas list in chat
[33056] Number of events should match the number of events in aggregates component
DisasterTypeComponent

[33028] Disaster type component elements should be visible
TimelineComponent

[33066] Timeline elements are displaying alert level correctly

For now the comments are mostly in regards to Ethiopia Malaria Dataset 75 and this will require additional work to setup Playwright to cover those failing tests.

### DIFF
--- a/tests/e2e/Pages/AggregatesComponent.ts
+++ b/tests/e2e/Pages/AggregatesComponent.ts
@@ -52,7 +52,8 @@ class AggregatesComponent extends DashboardPage {
 
   async aggregateComponentIsVisible() {
     await expect(this.aggregateSectionColumn).toBeVisible();
-    await expect(this.aggregateSectionColumn).toContainText('National View');
+    // We should remove this line as it is indicating text that is only visible and indifferrent to Uganda drought trigger and no trigger
+    // await expect(this.aggregateSectionColumn).toContainText('National View');
   }
 
   async aggregatesElementsDisplayedInNoTrigger(
@@ -108,6 +109,7 @@ class AggregatesComponent extends DashboardPage {
       hasText: 'Exposed population',
     });
     await this.page.locator('ion-backdrop').last().click();
+    // This test-id fails on ethiopia malaria
     await exposedPopulationLayer.getByTestId('aggregates-info-icon').click();
     const layerInfoTitle = await this.layerInfoPopoverTitle.textContent();
     const layerInfoContent = await this.layerInfoPopoverContent.textContent();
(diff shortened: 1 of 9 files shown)


jannisvisser:
@Piotrk39 to be honest I don't really get why you added these scenarios? At the moment there was no ambition to extend the number of scenarios, and it's totally expected that some tests would start failing. This is not related to the test your were creating on "set trigger" right?

Piotrk39:
> @Piotrk39 to be honest I don't really get why you added these scenarios? At the moment there was no ambition to extend the number of scenarios, and it's totally expected that some tests would start failing. This is not related to the test your were creating on "set trigger" right?

@jannisvisser those scenarios appear on their own once new dataset is created. So when we want to add typhoon that is only available in Philippines then we need to deal with all the other scenarios or skip them in some way. In this case I had to add both PHL and ETH.

jannisvisser:
this is not close to being ready and is not a priority

PR #2121, opened 2025-04-14
test: add integration tests for GET admin-areas and admin-areas/aggregates

gulfaraz:
## Describe your changes

[AB#35334](https://dev.azure.com/redcrossnl/IBF/_workitems/edit/35334)

This PR adds 2 simple integration tests on the /admin-areas and /admin-areas/aggregates endpoints.

## Notes for the reviewer

I stopped at the basic tests based on [this comment](https://dev.azure.com/redcrossnl/IBF/_workitems/edit/35225#14729479). It makes sense to investigate for potential refactors before adding more tests.

### DIFF
--- a/tests/integration/tests/admin-areas/admin-areas.test.ts
+++ b/tests/integration/tests/admin-areas/admin-areas.test.ts
@@ -291,6 +291,7 @@ export default function adminAreaTests() {
             return;
           }
           // REFACTOR: find out why MWI flash-flood responds differently, fix it, and remove this conditional assertion
+          // MWI flash flood returns event areas and not admin areas
           expect(feature.properties.name).toBeTruthy(); // the name should not be empty
           expect(feature.properties.adminLevel).toBe(adminLevel); // request and response admin levels should match
           expect(feature.properties.countryCodeISO3).toBe(countryCodeISO3); // request and response country codes should match
--- a/tests/integration/tests/admin-areas/admin-areas.test.ts
+++ b/tests/integration/tests/admin-areas/admin-areas.test.ts
@@ -16,12 +16,146 @@ export function getAdminAreas(
     .set('Authorization', `Bearer ${accessToken}`);
 }
 
-export default function adminAreaTests() {
-  const countryCodeISO3 = 'UGA';
-  const disasterType = DisasterType.Floods;
-  const adminLevel = AdminLevel.adminLevel3;
-  const scenario = FloodsScenario.Trigger;
+const assertions = [
+  {
+    countryCodeISO3: 'UGA',
+    disasterType: DisasterType.Floods,
+    adminLevel: AdminLevel.adminLevel2,
+    scenario: FloodsScenario.Trigger,
+    featureCount: 16,
+    placeCodeRegex: /^(UG|21UGA)/, // REFACTOR: set to /^UG/ after data is fixed
+  },
+  {
+    countryCodeISO3: 'UGA',
+    disasterType: DisasterType.Floods,
+    adminLevel: AdminLevel.adminLevel2,
+    scenario: FloodsScenario.Warning,
+    featureCount: 2,
+    placeCodeRegex: /^(UG|21UGA)/, // REFACTOR: set to /^UG/ after data is fixed
+  },
+  {
+    countryCodeISO3: 'UGA',
+    disasterType: DisasterType.Drought,
+    adminLevel: AdminLevel.adminLevel2,
+    scenario: FloodsScenario.Trigger,
+    featureCount: 4,
+    placeCodeRegex: /^(UG|21UGA)/, // REFACTOR: set to /^UG/ after data is fixed
+  },
+  {
+    countryCodeISO3: 'ZMB',
+    disasterType: DisasterType.Floods,
+    adminLevel: AdminLevel.adminLevel2,
+    scenario: FloodsScenario.Trigger,
+    featureCount: 14,
+    placeCodeRegex: /^ZM/,
+  },
+  {
+    countryCodeISO3: 'ZMB',
+    disasterType: DisasterType.Drought,
+    adminLevel: AdminLevel.adminLevel1,
+    scenario: FloodsScenario.Trigger,
+    featureCount: 10,
+    placeCodeRegex: /^ZM/,
+  },
+  {
+    countryCodeISO3: 'MWI',
+    disasterType: DisasterType.Floods,
+    adminLevel: AdminLevel.adminLevel3,
+    scenario: FloodsScenario.Trigger,
+    featureCount: 4,
+    placeCodeRegex: /^(?:MW|Karonga|Blantyre City|Rumphi)/, // REFACTOR: set to /^MW/ after data is fixed
+  },
+  {
+    countryCodeISO3: 'MWI',
+    disasterType: DisasterType.FlashFloods,
+    adminLevel: AdminLevel.adminLevel3,
+    scenario: FloodsScenario.Trigger,
+    featureCount: 3,
+    placeCodeRegex: /^(?:MW|Karonga|Blantyre City|Rumphi)/, // REFACTOR: set to /^MW/ after data is fixed
+  },
+  {
+    countryCodeISO3: 'SSD',
+    disasterType: DisasterType.Floods,
+    adminLevel: AdminLevel.adminLevel3,
+    scenario: FloodsScenario.Trigger,
+    featureCount: 1,
+    placeCodeRegex: /^SS/,
+  },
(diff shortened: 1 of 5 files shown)


tests/integration/tests/admin-areas/admin-areas.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  const scenario = FloodsScenario.Trigger;
+
+  describe('admin areas', () => {
+    let accessToken: string;
+
+    beforeAll(async () => {
+      accessToken = await getAccessToken();
+    });
+
+    it('should return list of admin areas on GET', async () => {
```
jannisvisser, line 32: This test is a good starting point and is good to have, so we could leave it at this. But I think the real value / addressing the real vulnerabilities comes here in testing/defining different scenarios. From the top of my head (not saying we should implement all of this at once): 
- Getting all areas in case of no-alert
- Getting all areas or just triggered ones based on disasterType.showOnlyTriggeredAreas (if that is still working as such actually, I think not)
- Getting areas with event-filter or not
- Getting only areas of a certain parentPlaceCode
- Getting only areas of the right alertLevel 

What do you think? Part of this also relates back to my comment in the as-is documentation, which e.g. may already point out that showOnlyTriggeredAreas distinction is no longer relevant. So maybe let that take precedence and then revisit this.

tests/integration/tests/admin-areas/admin-areas.test.ts
```diff
@@ -0,0 +1,58 @@
...
+      const adminAreas = await getAdminAreas(
+        countryCodeISO3,
+        disasterType,
+        adminLevel,
+        accessToken,
+      );
+
+      // Assert
+      expect(adminAreas.status).toBe(200);
+      expect(adminAreas.body.features.length).toBe(23);
```
jannisvisser, line 46: what is good practice about the expected number 23 here? Leave a comment explaining this? Assign a variable expectedNr=23 to at least give a bit more labeling? Work with a full .toMatchSnapshot assertion instead of just this?

tests/integration/tests/admin-areas/aggregates.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      const adminAreaAggregates = await getAdminAreaAggregates(
+        countryCodeISO3,
+        disasterType,
+        adminLevel,
+        accessToken,
+      );
+
+      // Assert
+      expect(adminAreaAggregates.status).toBe(200);
+      expect(adminAreaAggregates.body.length).toBe(138);
```
jannisvisser, line 48: same comment as above

tests/integration/tests/admin-areas/aggregates.test.ts
```diff
@@ -0,0 +1,56 @@
...
+  const scenario = FloodsScenario.Trigger;
+
+  describe('admin areas', () => {
+    let accessToken: string;
+
+    beforeAll(async () => {
+      accessToken = await getAccessToken();
+    });
+
+    it('should return list of admin area aggregates on GET', async () => {
```
jannisvisser, line 34: in principal same comment about scenarios as above, but at the same time maybe not as needed to duplicate all of that twice..

jannisvisser (commented):
Nice progress :) No clear approval/requested changes atm, but some questions/comments.

tests/integration/tests/admin-areas/admin-areas.test.ts
gulfaraz: @jannisvisser agreed

tests/integration/tests/admin-areas/admin-areas.test.ts
gulfaraz: @jannisvisser I tried `toMatchSnapshot` and found it convenient. But it created a file with 5319 lines for only 2 API calls `/admin-areas` and `/admin-areas/aggregates`. This option will exponentially bloat the repo.

I'll revert back to the hardcoded magic numbers with comments to explain how to interpret them. I'm open to alternatives.

jannisvisser: yeah that sounds about right. We could think of start using this for shorter responses or just to test specific parts of responses.

tests/integration/tests/admin-areas/admin-areas.test.ts
gulfaraz: https://github.com/rodekruis/IBF-system/blob/9e4c491cdfd36d97c0ca08ff8b8f506db3215d99/tests/integration/tests/admin-areas/admin-areas.test.ts#L278-L287

_Getting only areas of the right alertLevel_ fails for 2 scenarios

Should I create a bug or fix it within this test item? Also, take a look at the other REFACTOR comments in the file. @jannisvisser

jannisvisser: @gulfaraz I would say with a finding like this, at least do a little basic analysis to inform on the to-be-created-bug-item, with potentially the luck it's small enough you can fix it immediately (<2hours?)

jannisvisser (approved):
nice setup for the rest, approved except for the pending comment.

tests/integration/tests/admin-areas/admin-areas.test.ts
gulfaraz: @jannisvisser 

The 2 scenarios which caused this assertion to fail are,
1. UGA / flood / trigger
Screenshot 2025-04-18 at 10 08 02
2. MWI / flash-flood trigger
Screenshot 2025-04-18 at 10 07 37

These have multiple events with varying alert levels. We recently [decided](https://rodekruis.sharepoint.com/sites/510-CRAVK-510/_layouts/15/Doc.aspx?sourcedoc=%7B563bc23a-2574-430c-b8a9-541a8db3f345%7D&action=edit&wd=target%28IBF+Portal.one%7C97b7aca9-43d2-4874-8873-191e7d890663%2FDecisions%7C15219210-0b43-4ab6-b3e9-19e1226a238c%2F%29&wdorigin=703) that _Getting only areas of the right alertLevel_ is an event requirement. Our API fetches data for multiple events, hence the mixed alert levels.

I updated the test to assert this only if an eventName is specific.

https://github.com/rodekruis/IBF-system/blob/c68d3ad0f991b44707c96b9883a1623f5825c492/tests/integration/tests/admin-areas/admin-areas.test.ts#L278-L286

jannisvisser: nice. I was already suprised about this bug appearing, so good that it was an assertion-issue.

tests/integration/tests/admin-areas/admin-areas.test.ts
```diff
@@ -0,0 +1,301 @@
...
+              ),
+            ).toBeTruthy(); // all features should have the same alert level
+          }
+          if (
+            countryCodeISO3 === 'MWI' &&
+            disasterType === DisasterType.FlashFloods
+          ) {
+            return;
+          }
+          // REFACTOR: find out why MWI flash-flood responds differently, fix it, and remove this conditional assertion
```
jannisvisser, line 293: I guess this is because of the event-areas right? This endpoint returns event-areas in this case instead of admin-areas. (in National View, so without eventName). If not something that can be made actionable for now in the tests, let's at least update the comment to reflect this insight.

gulfaraz: Updated https://github.com/rodekruis/IBF-system/pull/2121/commits/ea51c3f33875f36177f3592119eae61796558068

I still don't understand why event areas do not contain a `name / admin level / country code`.

Feel free to update the comment with any more information that will help.

@jannisvisser

jannisvisser: This sounds like this can be improved, probably as part of AB#32715. Making the response format (more) aligned with the admin-area response format. In fact: 
- Country code could just be added (although not used), 
- name is now 'eventAreaName', but it doesn't seem to be used, so could just be changed to 'name' it seems. 
- only adminLevel is a problem, because event-areas really do not have an adminLevel. But the other 2 could maybe be changed in this PR.

jannisvisser: Updated https://github.com/rodekruis/IBF-system/pull/2121/commits/2833a1c1f3a90f7e9fed6b1a2c2e3976cab6b9cc

- I aligned the respone format of event-areas (more) to that of admin-areas. But left the refactor comment, that more can be done.
- Refined the test exception to only do so if no eventName (as only in National View this is different) and removed the countryCode. As the code exception on this is on disaster-type currently as well, not on country.

PR #2170, opened 2025-04-29
fix: foreign key constraint

gulfaraz:
## Describe your changes

Set [CASCADE DELETE](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-FK) between `event-place-code` and `admin-area` to resolve [AB#33928](https://dev.azure.com/redcrossnl/IBF/_workitems/edit/33928).

## Notes for the reviewer

See below my notes from [AB#35625](https://dev.azure.com/redcrossnl/IBF/_workitems/edit/35625),

We're trying to replace old (and incorrect) admin areas with new admin areas. However, the old admin areas are linked to event-place-codes. These connections are references to the rows in the admin-areas table.

To illustrate,

`admin-area` table has a row with `placeCode: UGA02`

`event-place-code` table has 2 rows with references `placeCode: UGA02`

If we attempt to delete `placeCode: UGA02` in the `admin-area` table.

Postgres first checks if any other tables reference the deleted data.
- If it finds no references, it deletes the row from the table `admin-area`.
- If it finds references, it looks for instructions on how to handle these references.
    - If it finds clear instructions, it performs them to clean up the references.
    - If it fails to find instructions, it throws an error because it doesn't know how to clean up these references.

The database doesn't allow references without sources. Otherwise, we lose [data integrity](https://en.wikipedia.org/wiki/Data_integrity#:~:text=Data%20integrity%20is%20the%20maintenance,%2C%20processes%2C%20or%20retrieves%20data.). For example, `event-place-code` has a row which references `placeCode: UGA02`. But `admin-area` does not have a row with `placeCode: UGA02`. Such a state is said to have no [data integrity](https://en.wikipedia.org/wiki/Data_integrity#:~:text=Data%20integrity%20is%20the%20maintenance,%2C%20processes%2C%20or%20retrieves%20data.) because it leads to partial or misleading information.

**Solution: Instruct the database to delete any referenced rows using [CASCADE DELETE](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-FK).**


jannisvisser:
@gulfaraz 
1. doesn't this clash with the requirement to keep data logs of all past events? So I'm not sure if this is an adequate solution..
2. am I correct to assess that this work immediately includes the remaining data work on UGA/MWI that was left pending before?

PR #2203, opened 2025-05-15
upload Eth static data

aklilu:
## Describe your changes
Upload csv files for the static layers


jannisvisser (approved):
looks good @aklilu! Thanks for the quick and thorough work. One thing I noticed is that 'name' and 'type' are often missing in the health_sites data, but I guess that's just how it is, so no blocker.

PR #2207, opened 2025-05-16
Feat.eth flash floods assets

jannisvisser:
## Describe your changes

- Configures and seeds static asset layers for roads/buildings/schools/health-sites/waterpoints
- Changed the geoserver layers roads_MWI and buildins_MWI to roads and buildings (becoming multi-country with parameters)
- Configures and mocks aggregate exposed asset layers per admin-area for each
- Mocks exposed assets + needed changes to make this possible.
- Related pipeline change MWI flash-floods: https://github.com/rodekruis/IBF-flash-flood-pipeline/pull/73


interfaces/IBF-dashboard/src/app/types/ibf-layer.ts
```diff
@@ -56,6 +56,7 @@ export enum IbfLayerType {
   point = 'point',
   shape = 'shape',
   wms = 'wms',
+  line = 'line',
```
gulfaraz, line 59: Roads and buildings load WMS tiles on the frontend. Thus making their layer type WMS. Introducing a new LINE type to behave exactly as the WMS layer type does not make sense (to me).

It also leads to unnecessary extra code like,

```typescript
[IbfLayerType.wms, IbfLayerType.line].includes(newLayer.type)

case IbfLayerType.wms || IbfLayerType.line:

if (layer.type === IbfLayerType.wms || layer.type === IbfLayerType.line) {
```

By definition, if the layer loads WMS tiles, its layer type is line. If you want to disambiguate behaviour for other reasons as here,
```typescript
      const nameLeadTimePart = leadTime ? `_${leadTime}` : '';
      const nameCountryPart =
        layer.type === IbfLayerType.line
          ? ''
          : `_${this.country.countryCodeISO3}`;
      const name = `ibf-system:${layer.name}${nameLeadTimePart}${nameCountryPart}`;
```

Then use a different parameter like layer name. In this case, it also doesn't make sense why roads and buildings are not country-specific.

services/API-service/src/api/lines-data/lines-data.service.ts
```diff
@@ -102,6 +102,7 @@ export class LinesDataService {
 
       const validationError = await validate(linesDto);
       if (validationError.length > 0) {
+        console.log('Seed line data validation error: ', validationError);
```
gulfaraz, line 105: Use `this.logger.log` instead of `console.log`. See [SeedAdminAreaData](https://github.com/rodekruis/IBF-system/blob/32a7ac8a02f0794112b4ec588258a525331c5c72/services/API-service/src/scripts/seed-admin-area-data.ts#L21) for reference.

services/API-service/src/api/point-data/point-data.service.ts
```diff
@@ -219,6 +219,7 @@ export class PointDataService {
 
       const validationError = await validate(pointDto);
       if (validationError.length > 0) {
+        console.log('Seed point data validation error: ', validationError);
```
gulfaraz, line 222: Use `this.logger.log` instead of `console.log`. See [SeedAdminAreaData](https://github.com/rodekruis/IBF-system/blob/32a7ac8a02f0794112b4ec588258a525331c5c72/services/API-service/src/scripts/seed-admin-area-data.ts#L21) for reference.

services/API-service/src/scripts/seed-point-data.ts
```diff
@@ -72,8 +77,13 @@ export class SeedPointData implements InterfaceScript {
         countryCodeISO3,
         pointDtos,
       );
-    } catch {
-      return Promise.resolve();
+    } catch (error) {
+      // If validation or upload fails, then log and throw error
+      console.error('Error seeding point data:', error);
```
gulfaraz, line 82: Use `this.logger.log` instead of `console.log`. See [SeedAdminAreaData](https://github.com/rodekruis/IBF-system/blob/32a7ac8a02f0794112b4ec588258a525331c5c72/services/API-service/src/scripts/seed-admin-area-data.ts#L21) for reference.

gulfaraz (changes requested):
Buildings layer loads as expected,
Screenshot 2025-05-21 at 12 14 54

Roads layer loads as expected,
Screenshot 2025-05-21 at 12 14 48

Schools layer loads as expected,
Screenshot 2025-05-21 at 12 19 35

Health sites layer loads as expected,
Screenshot 2025-05-21 at 12 19 42

Waterpoints layer loads as expected,
Screenshot 2025-05-21 at 12 26 35

Second? Waterpoints layer loads as expected,
Screenshot 2025-05-21 at 12 26 45

Do we expect to see 2 Waterpoints layers?

I kindly request that we avoid adding a new LINE layer type - unless it loads line information in line format. This PR introduces a new layer type (LINE) that behaves identically to an existing layer type (WMS).

interfaces/IBF-dashboard/src/app/types/ibf-layer.ts
jannisvisser: I introduced this new type specifically for this last use case, instead of having to explicitly mention 2 layer names here, which feels less robust. But if you prefer I can instead create a const or enum that specifies which layers are in this group of not-country-specific-geoserver-layers.

This whole thing was indeed in turn needed because I decided to make the 'roads' and 'buildings' Geoserver layers generic, instead of having to create a new layer for each (and keep doing so for new countries in the futrure). This is meant to make it more scalable. This cannot be (easily) applied to other Geoserver layers beause they are based on files not on a database view. Do you disagree with this decision?

jannisvisser: no we pass countryCodeISO3 in 'viewparams' [here](https://github.com/rodekruis/IBF-system/pull/2207/files#diff-50a4787f0cc34ca7a9bd50c25ce9cfcc581518bb852ecc3bcb2ebb6b4f6aca73), which is read by the SQL view in Geoserver [here](https://github.com/rodekruis/IBF-system/pull/2207/files#diff-74f6357413e23d680820233a71126c3536959e1ae1c5d7cda045e1450987176e).

jannisvisser: See https://github.com/rodekruis/IBF-system/pull/2207/commits/56fe8a8b57c8174ae7569ebbaca42726e7e7074d.

There are also references to type=line in the e2e code, which I will address in the separate test PR.

So if you review this commit, this PR should be mergeable.

gulfaraz (approved):
In trigger scenario,

Waterpoints layer,
Screenshot 2025-05-22 at 10 34 40

Schools layer,
Screenshot 2025-05-22 at 10 34 58

Health sites layer,
Screenshot 2025-05-22 at 10 35 12

Roads layer,
Screenshot 2025-05-22 at 10 35 33

Buildings layer,
Screenshot 2025-05-22 at 10 35 44

PR #2216, opened 2025-05-20
refactor: remove event areas

gulfaraz:
## Describe your changes

This PR includes,
1. Event areas should be dynamically stitched based on underlying admin areas. The core changes are in getEventAdminAreas and getEventAdminArea.
2. Remove the functions addOrUpdateEventAreas, getEventAreas and the entity EventAreaEntity, as they are no longer needed.

Good to know that I tried 2 alternative approaches,
1. In getEventAdminAreas, I tried stitching the admin areas incrementally as I looped through the admin areas. Similar to how I reduce the main exposure indicator to calculate the total per event. This approach worked, but the performance was slow, about 7 times slower on average. For instance, ~20ms vs ~140ms. Likely because the union function is expensive, so we only call it (at most) once per event.
2. This is the most preferred option. I verified that it is possible to use [ST_Union](https://postgis.net/docs/ST_Union.html) to stitch the admin areas within Postgres. However, I couldn’t rewrite our DB query to get the same result. We build our query with many in-code clauses instead of SQL joins. Major refactoring is needed to make this approach work. Instead, I reverted to an in-code aggregation, thus making the future refactoring more tedious.

I failed to make this work across countries and hazards. I ran into dependencies which break features like trigger layer, GloFAS stations, set trigger, and validation errors. My attempts to fix these quickly expanded to major refactors.

## Notes for the reviewer

1. Dropping a TABLE with migration files is an [open discussion](https://github.com/typeorm/typeorm/issues/7814) in TypeORM. For convenience, I removed the event-area CREATE TABLE migration file and added [a task](https://dev.azure.com/redcrossnl/IBF/_workitems/edit/36152) to the release item to drop the table manually in demo and production.
2. I added stronger typing to using geometry types from geojson. This led to validation errors. Then I found that we faced similar errors before leading to [this config](https://github.com/rodekruis/IBF-system/blob/2807f3bb362ba62ae1b530466d118f6a2f02f676/services/API-service/src/main.ts#L57). I applied the same fix with FIX and TODO comments to improve on this.


services/API-service/src/api/admin-area/admin-area.service.ts
```diff
@@ -489,10 +470,22 @@ export class AdminAreaService {
...
+    if (disasterType === DisasterType.FlashFloods && !eventName) {
+      // TODO: use IF admin level is national view (or less than default admin level ?)
+      const eventAdminAreas = this.eventAreaService.getEventAdminAreas(
+        adminAreas,
+        indicator,
+      );
+
+      if (eventAdminAreas.features.length > 0) {
+        return eventAdminAreas;
+      }
```
jannisvisser, line 486: so if no event-areas are found (if there are no event? so on no-alert) it reverts to normal admin-area returning? This would be a functional change for MWI flash-floods, as currently it shows all potential-event-areas/adm2-areas in no-alert state instead of now all adm3-areas?

services/API-service/src/api/admin-area/services/event-area.service.ts
```diff
@@ -237,4 +134,77 @@ export class EventAreaService {
...
+    }
+
+    if (properties.alertLevel == AlertLevel.NONE) {
+      // return null to exclude no alert events
+      // getAdminAreas will fallback to admin areas if no alert event is found
+      return null;
+    }
+
+    if (adminAreas.length > 1) {
+      return union(featureCollection(adminAreas), { properties });
```
jannisvisser, line 203: awesome :)

services/API-service/src/api/admin-area/admin-area.service.ts
gulfaraz: > so if no event-areas are found (if there are no event? so on no-alert) it reverts to normal admin-area returning?

yes

> This would be a functional change for MWI flash-floods, as currently it shows all potential-event-areas/adm2-areas in no-alert state instead of now all adm3-areas?

I see admin level 2 areas on test (v0.317.0) and admin level 3 areas on this branch.

Admin level 2 areas shown on test (v0.317.0),
Screenshot 2025-05-21 at 10 04 54

Admin level 3 areas shown on this branch,
Screenshot 2025-05-21 at 10 04 48

So, yes, this is a functional change. Would you like us to show admin level 2 areas? @jannisvisser

services/API-service/src/api/admin-area/admin-area.service.ts
```diff
@@ -489,10 +470,22 @@ export class AdminAreaService {
     const highestAlertLevels =
       this.eventService.getHighestAlertLevelPerEvent(adminAreas);
     adminAreas = adminAreas.filter(
-      (area) =>
-        area.alertLevel === highestAlertLevels[area.eventName || 'unknown'],
+      ({ alertLevel, eventName }) =>
+        alertLevel === highestAlertLevels[eventName || 'unknown'],
     );
 
+    if (disasterType === DisasterType.FlashFloods && !eventName) {
```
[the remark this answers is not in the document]
gulfaraz: @jannisvisser The stitching worked for all countries and hazards. I didn't note any technical limitations. I ran into _seemingly_ unrelated issues (example, glofas stations, trigger layer, data validation errors), all of which I could not check manually. Thus lowering my confidence in the final result. In the end, I opted to limit the scope to flash floods, as it is the only hazard that uses event areas.

To check for other hazards, you can replace,
```typescript
if (disasterType === DisasterType.FlashFloods && !eventName) {
```
with
```typescript
if (!eventName) {
```

services/API-service/migration/1695644873629-eventAreas.ts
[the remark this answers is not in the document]
gulfaraz: I created a migration script in https://github.com/rodekruis/IBF-system/pull/2216/commits/571e327c47a1169cfa1ec8de2828871918bba590 as advised.

> I don't think this is good practice around migration script.

Would it be possible to share your source for this advice? I couldn't find anything conclusive on the web.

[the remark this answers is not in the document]
jannisvisser: didn't have a specific source in mind, but CoPilot says this on request:
----------------------
Database Migration Best Practices: To Modify or Not to Modify
When dealing with database migrations, a common question is whether to modify/delete old migration scripts or keep them as a complete history. Here's the industry consensus:

Best Practice: Keep Complete Migration History
Generally, you should not delete or modify old migrations that have been applied to any environment. Here's why:

Database State Tracking: Migrations represent the historical evolution of your database schema
Version Control: They serve as documentation of how the database reached its current state
Team Collaboration: Other developers rely on the migration history
Production Safety: Deleting migrations can break deployments to production
Sources Supporting This Approach
Several database migration frameworks explicitly follow this philosophy:

Flyway Documentation
Flyway, a popular migration tool, states:

"Migrations that have already been applied to a database should never be modified. Migrations are intended to be immutable."

Entity Framework Core (Microsoft)
Microsoft's EF Core documentation recommends:

"Avoid modifying or removing a migration after it's been applied to a production database. Doing so can corrupt your database or require manual intervention to fix."

Ruby on Rails Guides
Rails migrations documentation advises:

"In general, editing existing migrations is not a good idea. You will be creating extra work for yourself and your co-workers and cause major headaches if the existing version of the migration has already been run on production machines."

How to Handle Unneeded Tables
If you have tables created by old migrations that are no longer needed:

Create a new migration to drop the table/column
Keep the original migration that created it
Document the change in comments or migration naming
This approach maintains the historical record while still cleaning up your schema.

Exceptions
The only reasonable times to modify migrations that haven't been applied to production:

During local development before sharing with other team members
When fixing migrations that haven't been applied to any shared environment yet
When working on a completely new branch with a fresh database

services/API-service/src/api/admin-area/admin-area.service.ts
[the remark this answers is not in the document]
jannisvisser: see new commit which seems to successfully use event-areas for floods also

services/API-service/src/api/admin-area/admin-area.service.ts
jannisvisser: no, this is because in a previous PBI I "unsimplified" only the Dire Dawa adm3-areas. At that time they were never in the same view as other adm3-areas, because you would only see adm3-areas in Event View, and now you also see them in National View.

gulfaraz: Right, so we are mixing simplified and unsimplified GeoJSONs. In that case, the overlap is logical.

jannisvisser: this will be resolved in separate PR

PR #2217, opened 2025-05-21
Feat.event areas drought

jannisvisser:
## Describe your changes

- adds drought as event-area disaster-type
- sets up a const to define which disaster-type have event-areas
- small fix needed on eventName for drought

## Notes for the reviewer

1. I checked trigger and no-trigger for each disaster-type
2. This does not include performance check on drought yet. But that can be separate. The refactor (definining a const) in this PR is needed anyway, we might just take drought ouf of the const array.

### DIFF
--- a/services/API-service/src/api/admin-area/services/event-area.service.ts
+++ b/services/API-service/src/api/admin-area/services/event-area.service.ts
@@ -151,7 +151,7 @@ export class EventAreaService {
 
         events.push({
           placeCode: eventName,
-          name: eventName,
+          name: eventName?.split('_')[0],
           eventName,
           countryCodeISO3,
           [indicator]: indicatorValue ?? 0,
--- a/services/API-service/src/api/admin-area-dynamic-data/admin-area-dynamic-data.service.ts
+++ b/services/API-service/src/api/admin-area-dynamic-data/admin-area-dynamic-data.service.ts
@@ -14,6 +14,7 @@ import {
 import { DisasterTypeGeoServerMapper } from '../../scripts/disaster-type-geoserver-file.mapper';
 import { HelperService } from '../../shared/helper.service';
 import { EventAreaService } from '../admin-area/services/event-area.service';
+import { EVENT_AREA_DISASTER_TYPES } from '../disaster-type/const/disaster-type.const';
 import { DisasterType } from '../disaster-type/disaster-type.enum';
 import { AlertLevel } from '../event/enum/alert-level.enum';
 import { EventService } from '../event/event.service';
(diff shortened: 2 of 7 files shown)


services/API-service/src/api/admin-area/services/event-area.service.ts
```diff
@@ -151,7 +151,7 @@ export class EventAreaService {
 
         events.push({
           placeCode: eventName,
-          name: eventName,
+          name: eventName?.split('_')[0],
```
gulfaraz, line 154: Is it possible to avoid this split across the code base? I see it appears in,
1. [mock-helper.service](https://github.com/rodekruis/IBF-system/blob/7bb895abf0ea39ef7b000a8bd64686a302a1bdbe/services/API-service/src/scripts/mock-helper.service.ts#L278-L279)
2. [alert-area.service](https://github.com/rodekruis/IBF-system/blob/7bb895abf0ea39ef7b000a8bd64686a302a1bdbe/interfaces/IBF-dashboard/src/app/services/alert-area.service.ts#L196)
3. [admin-level.component.html](https://github.com/rodekruis/IBF-system/blob/7bb895abf0ea39ef7b000a8bd64686a302a1bdbe/interfaces/IBF-dashboard/src/app/components/admin-level/admin-level.component.html#L43)
4. [notification-content.service](https://github.com/rodekruis/IBF-system/blob/7bb895abf0ea39ef7b000a8bd64686a302a1bdbe/services/API-service/src/api/notification/notification-content/notification-content.service.ts#L110)
5. and now [event-area.service](https://github.com/rodekruis/IBF-system/blob/7bb895abf0ea39ef7b000a8bd64686a302a1bdbe/services/API-service/src/api/admin-area/services/event-area.service.ts#L154)

To localize this logic, I suggest splitting the eventName only once in `/process-events`. The limitation that event names should not contain underscores should also be explicit in the wiki.

My suggestion can also be a separate refactor because it is pre-existing.

gulfaraz (approved):
Nice!

services/API-service/src/api/admin-area/services/event-area.service.ts
jannisvisser: yeah it even has a REFACTOR note already in [one of these places](https://github.com/rodekruis/IBF-system/blob/7bb895abf0ea39ef7b000a8bd64686a302a1bdbe/services/API-service/src/api/notification/notification-content/notification-content.service.ts#L108), so agree. 

In practice, the underscore-splitting (because of a needed part behind the underscore) is only needed in [one place in mock](https://github.com/rodekruis/IBF-system/blob/master/services/API-service/src/scripts/mock-helper.service.ts#L277), which should be solvable in another way. If so, we can lose this splitting altogether (and change the drought mock event names).

PR #2218, opened 2025-05-22
fix: event area aggregates should show 0 instead of NA

gulfaraz:
## Describe your changes

According to bug [AB#35929](https://dev.azure.com/redcrossnl/IBF/_workitems/edit/35929), event area aggregates should show 0 instead of NA.

The dashboard used to show the aggregate value received from the API, or fall back to NA to indicate that the value was unavailable.

Since event areas,
1. `/admin-areas/aggregates` API does not return a list of admin areas.
2. When events are found, the API returns a list of aggregate values. The dashboard shows the values (as expected).
3. When no events are found, the API returns an empty list. The dashboard falls back to NA (because no values in the API response).

I considered the following solutions,
1. The API should return a list of aggregate values, with aggregate value 0. The aggregates API response requires 3 properties,
```json
{
    "placeCode": "UG06023",
    "indicator": "population_affected",
    "value": 43
}
```
The values for `indicator` and `value` are clear. The value for `placeCode` is unclear. I expect `placeCode` to return a valid `placeCode` value. When I checked the API response for the case when events are found,
```json
{
    "placeCode": "Rumphi",
    "indicator": "population_affected",
    "value": 434
}
```
The `placeCode` value is the event name. _Let's assume this makes sense._ When there are no events, I don't have an event name to fill the `placeCode` value. The dead end: either remove `placeCode` from the aggregates response _or_ fill the requirements gap introduced by event areas.

2. The expected behaviour is relatively easily achieved by making the dashboard fallback to 0 instead of NA. This PR applies this solution. This solution assumes that the NA fallback is not a requirement but an implementation detail.

### DIFF
--- a/interfaces/IBF-dashboard/src/app/components/aggregates/aggregates.component.html
+++ b/interfaces/IBF-dashboard/src/app/components/aggregates/aggregates.component.html
@@ -113,7 +113,7 @@
                 </ion-label>
               } @else {
                 <ion-label class="ion-text-right m-0" data-test="aggregate-na">
-                  {{ 'aggregates-component.not-applicable' | translate }}
+                  {{ 'aggregates-component.default-value' | translate }}
                 </ion-label>
               }
             </ion-item>
--- a/interfaces/IBF-dashboard/src/assets/i18n/en.json
+++ b/interfaces/IBF-dashboard/src/assets/i18n/en.json
@@ -1,8 +1,6 @@
 {
   "app-name": "IBF-dashboard",
-  "common": {
-    "version": "Version"
-  },
+  "common": { "version": "Version" },
   "export-view": {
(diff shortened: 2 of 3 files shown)


jannisvisser:
@gulfaraz I think your solution doesn't fully cover it, because the aggregate should not always be 0. E.g. for a static indicato rlike 'Total population' (UGA floods), it should not say 0. But I can also imagine the theoretical possibility of dynamic exposure indicator having positive values, even though there is no event.

image

gulfaraz:
@jannisvisser FYI 🐞 [AB#36249](https://dev.azure.com/redcrossnl/IBF/_workitems/edit/36249)

jannisvisser (approved):

PR #2231, opened 2025-05-28
fix: hack to avoid duplicate waterpoints layers

gulfaraz:
## Describe your changes

This PR includes a hack to avoid duplicate waterpoints layers. [AB#36188](https://dev.azure.com/redcrossnl/IBF/_workitems/edit/36188)

## Notes for the reviewer

Please read notes in code comment.

### DIFF
--- a/interfaces/IBF-dashboard/src/app/services/map.service.ts
+++ b/interfaces/IBF-dashboard/src/app/services/map.service.ts
@@ -295,27 +295,31 @@ export class MapService {
     const isLoading = waterPoints ? false : true;
 
     // HACK: conditionally add waterpoints layer because it has a loading logic (see loadWaterPointsLayer)
-    // which adds the layer to the map asynchronously in portals where it is not configured
-    // multiple factors are at play here:
-    // 1. waterpoints and waterpoints_internal layers are used to identify the same layer
-    //    which breaks the technical constraint of using unique layer names
-    // 2. loadWaterPointsLayer is calls addWaterPointsLayer 2 times, the second call is async
-    //    - this acts up when the portals are switched quickly
-    //    - user navigation is too slow to be affected by the async call
-    // 3. the subscription hell triggers multiple calls to /metadata/layers
+    // which adds the layer to the map asynchronously in disaster types where it is not configured
+    // there are a couple of factors are at play here:
+    // 1. loadWaterPointsLayer is calls addWaterPointsLayer 2 times, the second call is async
+    //    - this acts up when the disaster types are switched quickly
+    //    - user navigation is usually too slow to be affected by the async call
+    // 2. the subscription hell triggers multiple calls to /metadata/layers
     //    first with old country and disaster type which starts an async addLayer
     //    then with new country and disaster type which starts another async addLayer
     //    the map layers are reset by the time the old async addLayer is resolved
     //    this causes the waterpoints layer to be added to the layers array
     //    if the layers shared the same name, the layer would be replaced
-    // Using the knowledge that the waterpoints layer is always in the layers array before we receive waterpoints data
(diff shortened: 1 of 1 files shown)


interfaces/IBF-dashboard/src/app/services/map.service.ts
```diff
@@ -293,21 +293,46 @@ export class MapService {
...
-      viewCenter: false,
-      order: 2,
-      isLoading,
-    });
+
+    // HACK: conditionally add waterpoints layer because it has a loading logic (see loadWaterPointsLayer)
+    // which adds the layer to the map asynchronously in portals where it is not configured
+    // multiple factors are at play here:
+    // 1. waterpoints and waterpoints_internal layers are used to identify the same layer
+    //    which breaks the technical constraint of using unique layer names
```
jannisvisser, line 301: I would still suggest to remove this comment. You are right in saying that IF it was also called 'waterpoints', this particular problem would not have appeared. But that is just 'luck' in the end. Say that 'waterpoints_internal' would not have been present in flash-floods, then still the same bug would happen, in the sense that 'waterpoints' would show up in flash-floods when moving quickly from floods to flash-floods. Likewise, I imagine that previously this bug would occur when switching from floods/drought to malaria.

Don't get me wrong. The '2 layers with same label' situation should also be cleared up (new PBI?), but I would suggest to leave it out here for conciseness/clearness of the comment.

interfaces/IBF-dashboard/src/app/services/map.service.ts
```diff
@@ -293,21 +293,46 @@ export class MapService {
...
-      ),
-      show: true,
-      data: waterPoints,
-      viewCenter: false,
-      order: 2,
-      isLoading,
-    });
+
+    // HACK: conditionally add waterpoints layer because it has a loading logic (see loadWaterPointsLayer)
+    // which adds the layer to the map asynchronously in portals where it is not configured
```
jannisvisser, line 298: should 'in portals' here specifically be 'in disaster-types'?

interfaces/IBF-dashboard/src/app/services/map.service.ts
```diff
@@ -293,21 +293,46 @@ export class MapService {
...
+    //    if the layers shared the same name, the layer would be replaced
+    // Using the knowledge that the waterpoints layer is always in the layers array before we receive waterpoints data
+    // we can restrict calling addLayer to the following conditions:
+    // CONDITION 1: add layer if loading and is not already present
+    // CONDITION 2: add layer if not loading and is already present
+    // the condition in the if statement is a simplification of the above two conditions
+    const hasWaterPointsLayer = this.layers.some(
+      ({ name }) => name === IbfLayerName.waterpoints,
+    );
+    if (isLoading !== hasWaterPointsLayer) {
```
jannisvisser, line 319: leaving up to you, but I can imagine for future/other developers it might be clearer to write out the 2 conditions separately. At the same time (and either way) that gives room to add the consideration on what part of this is for performance only, and what part is really needed for solving the bug?

jannisvisser (approved):
Solution looks good as discussed. (Although I cannot functionally test anything as I can not reproduce the issue). Comments are just on the in-code-comments.

interfaces/IBF-dashboard/src/app/services/map.service.ts
gulfaraz: @jannisvisser I renamed portals to disaster types

interfaces/IBF-dashboard/src/app/services/map.service.ts
gulfaraz: @jannisvisser I added a comment to clarify this.

PR #2700, opened 2026-04-16
feat: old login

gulfaraz:
[AB#41063](https://dev.azure.com/redcrossnl/9e7ca3cc-bb97-4471-a25c-fd9787af0fe8/_workitems/edit/41063)

## Describe your changes

1. Restore the password-login page at `/login`
2. Restore change password (frontend and backend)
3. Restore forgot password (frontend only)
4. Open `GET /country` as required by the old login page design
5. Hide passwordless-login page at `/login/backdoor`
6. Fix startup error introduced in https://github.com/rodekruis/IBF-system/pull/2590
7. Show _change password_ instead of _the manage page_ in the user menu
8. Minimal changes to tests to merge the PR
9. Remove the password hashing hack in user.entity.ts

The `/login/backdoor` and `/manage` pages are **only** accessible by entering the direct URL. They should not be accessible by normal user interaction.

### DIFF
--- a/interfaces/IBF-dashboard/src/app/components/backdoor-form/backdoor-form.component.ts
+++ b/interfaces/IBF-dashboard/src/app/components/backdoor-form/backdoor-form.component.ts
@@ -1,0 +1,177 @@
+import { Component, OnInit, ViewChild } from '@angular/core';
+import { NgForm } from '@angular/forms';
+import { ActivatedRoute, Router } from '@angular/router';
+import { IonInput } from '@ionic/angular';
+import { TranslateService } from '@ngx-translate/core';
+import {
+  AnalyticsEvent,
+  AnalyticsPage,
+} from 'src/app/analytics/analytics.enum';
+import { AnalyticsService } from 'src/app/analytics/analytics.service';
+import { AuthService } from 'src/app/auth/auth.service';
+import { LoginRequest, MessageResponse } from 'src/app/types/api';
+
+@Component({
+  selector: 'app-backdoor-form',
+  templateUrl: './backdoor-form.component.html',
+  styleUrls: ['./backdoor-form.component.scss'],
+  standalone: false,
+})
+export class BackdoorFormComponent implements OnInit {
+  @ViewChild('loginForm')
+  public loginForm: NgForm;
+
+  @ViewChild('emailInput', { static: false })
+  public emailInput: IonInput;
+
+  @ViewChild('codeInput', { static: false })
+  public codeInput: IonInput;
+
+  public formState: 'code' | 'email' = 'email';
+  public model = { email: '', code: '' };
+  public resendCodeDisabled = false;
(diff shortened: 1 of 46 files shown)


interfaces/IBF-dashboard/src/app/components/user-state-menu/user-state-menu.component.html
```diff
@@ -5,9 +5,14 @@
...
-  <ion-item [button]="true" [detail]="false" routerLink="manage">{{
-    'user-state.manage' | translate
+  <ion-item [button]="true" [detail]="false" (click)="presentPopover()">{{
+    'user-state.password-change' | translate
   }}</ion-item>
+  <!--
+  <ion-item [button]="true" [detail]="false" routerLink="manage">{{
+      'user-state.manage' | translate
+    }}</ion-item>
+  -->
```
jannisvisser, line 16: maybe include a short reason for this?

tests/e2e/global-setup.ts
```diff
@@ -32,10 +32,10 @@ async function globalSetup(config: FullConfig) {
       // launch browser
       const browser = await chromium.launch();
       const page = await browser.newPage();
-      await page.goto(baseURL!);
+      await page.goto(`${baseURL!}/login/backdoor`);
```
jannisvisser, line 35: does this mean all e2e tests are using the backdoor login? Why? Is that easier?

interfaces/IBF-dashboard/src/app/components/backdoor-form/backdoor-form.component.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    this.resendCodeDisabled = true;
+
+    const loginRequest: LoginRequest = { email };
+
+    if (code) {
+      loginRequest.code = code;
+    }
+
+    this.authService.backdoor(loginRequest).subscribe({
```
jannisvisser, line 111: I notice that when I use my personal non-redcross email, and then fill in the code from dev tools, it gives a 401 and then reverts to /login instead of to /login/backdoor. Even though the path is not likely to be used, maybe good to fix if easy?

jannisvisser (approved):
Solid. Approved, but a few small comments/questions. Up to you to process or not.

tests/e2e/global-setup.ts
gulfaraz: Yes, there are significant changes to the e2e tests based on passwordless login. We can redo the e2e tests or simply delete them but this approach was least effort.

interfaces/IBF-dashboard/src/app/components/backdoor-form/backdoor-form.component.ts
gulfaraz: Good catch. I'll timebox this to an hour.

